PR #6, opened 2019-08-01
Update faq.md

denis-yuen:
Also demoing how/where we will work on doc updates for 1.7.0


docs/faq.md
```diff
@@ -11,6 +11,18 @@ We can also provision input and output files using HTTP, FTP, and S3. We also ha
 
 Typically, we test running tools in Ubuntu Linux 16.04 LTS on VMs in [OpenStack](https://www.openstack.org/) with 8 vCPUs and 96 GB of RAM and above. We've also begun testing on Ubuntu 18.04 LTS and so far it's been successful. If you are only listing and editing tools, we have achieved success with much lower system requirements. However, launching tools will have higher system requirements dependent on the specific tool. Consult a tool's README or CWL/WDL description when in doubt.
 
+## The CLI is failing with Java 8 
+
+If you run into:
```
coverbeck: Having a fragment start with `If` and end with a colon feels wrong, although I can't find a formal rule on it. I guess I'm expecting a `then` (or an implied one) after I see an `if`.

Maybe `If you see the following error when running the Dockstore CLI, you need to upgrade your Java version:`.

PR #10, opened 2019-08-09
Feature/bosc content

agduncan94:
Added a page for talk content in the Extras section. Currently this just includes BOSC 2019, but any other talks/posters are welcome.


agduncan94:
Screenshot from 2019-08-09 14-24-46

docs/posters-and-talks.rst
```diff
@@ -0,0 +1,34 @@
...
+
+
+2019
+----
+
+BOSC 2019
+^^^^^^^^^
+Dockstore 2.0: Enhancing a community platform for sharing cloud-agnostic research tools
+
+- :download:`Poster </_static/2019-BOSC_conference_poster.pdf>`
```
denis-yuen: Should include citation info via DOI to allow people to cite this presentation. 
This should be auto-generated on the F1000 site ala the other issue with links

agduncan94:
Screenshot from 2019-08-12 09-23-52

agduncan94:
@denis-yuen Do you mean put it in only the first release (ex. 1.7.0) or put it in every release?

I think I'd rather just keep it in the repo directly but no strong feelings.

denis-yuen:
> @denis-yuen Do you mean put it in only the first release (ex. 1.7.0) or put it in every release?
> 
> I think I'd rather just keep it in the repo directly but no strong feelings.

Just the first time it appears. 
Feels a bit weird since one of the posters is 20MB, but re-reading 
https://docs.microsoft.com/en-us/azure/devops/repos/git/manage-large-files?view=azure-devops

I think it passes because while it is an "output",  its neither particularly large and is not frequently updated.

PR #11, opened 2019-08-09
Feature/fix broken external links

garyluu:
Fix broken and redirected external links.  It appears the linkcheck doesn't really do internal links that well.


docs/faq.md
```diff
@@ -20,18 +20,18 @@ Error: A JNI error has occurred, please check your installation and try again
...
+You will need to upgrade from Java 8 to use the CLI.
 
 ## There are too many versions of my tool, how do I delete some?
 
 Versions of your tool for most tools are harvested from the list of Tags for an image on Quay.io, [as an example](https://quay.io/repository/pancancer/pcawg-bwa-mem-workflow?tab=tags). If you have the right permissions, you can delete some and then refresh a tool on Dockstore to clean-up.
 
 ## How do I cite Dockstore?
 
-For citing Dockstore as a paper, take a look at our [F1000 paper](http://dx.doi.org/10.12688/f1000research.10137.1).
+For citing Dockstore as a paper, take a look at our [F1000 paper](https://f1000research.com/articles/6-52/v1).
```
coverbeck: Why did you change the host as well? Aren't DOI's supposed to more permanent?

garyluu: it redirects to this

denis-yuen: Think @coverbeck is right though, the DOI is permanent but what it points to is not guaranteed.
That's why the auto-generated text in the article says this 

> How to cite: O'Connor BD, Yuen D, Chung V et al. The Dockstore: enabling modular, community-focused sharing of Docker-based genomics tools and workflows [version 1; peer review: 2 approved]. F1000Research 2017, 6:52 > > (https://doi.org/10.12688/f1000research.10137.1)

garyluu: So you want to just ignore redirects from doi.org then?

PR #31, opened 2019-09-26
Document Dockstore GitHub App

coverbeck:
dockstore/dockstore#2824

This is pretty thin, but Natalie's getting started with services already covers a lot of this, so I just linked to it rather than duplicating. I could go any way on whether it's worth going with this, improving it some more, or just dropping it entirely.

Visible here: https://beta.docs.dockstore.org/en/feature-2824-githubapps/getting-started/github-apps.html

Notes:

* The links from this page do not work, but I discovered a generic problem, which I logged separately, dockstore/dockstore#2925 -- we should fix it the right way everywhere, but I don't know what the right way is, so held off.
* Also did not add the discourse id -- waiting for consensus as to whether we should go ahead with this.


agduncan94:
I think a github apps focused page makes sense, since once it is implemented for wf too we probably want a general page on what it is for. Also it is short and simple so we could link to this as more of a landing page for github apps docs, and then people could click into more specific examples with services.

garyluu:
Having trouble with internal links? :smile:

coverbeck:
> Having trouble with internal links? 😄

Not any more!

PR #34, opened 2019-10-03
redo wdl best practices

NatalieEO:
The example being used previously is not working. I spent some time trying to fix it, but thought it was better to use same example from getting started tutorial instead

https://docs.dockstore.org/en/wdl-1.0/


docs/advanced-topics/best-practices/wdl-best-practices.rst
```diff
@@ -1,7 +1,14 @@
 WDL Best Practices
 ==================
 
-.. include:: best-practices-intro.rst
+Best Practices
+--------------
+
+Following are some best practices for creating tools. Our intention is
+that this document will evolve as descriptor languages evolve so feel
```
coverbeck: You have the word `evolve` twice in the sentence and "Our intention is..." seems unnecessary. I feel like this could read a little better, but no concrete suggestions.

NatalieEO: reworded

NatalieEO:
Ah I see. Good point. I left it as a tool since that is how it is used in the getting started wdl which does mention that wdl doesn't have the concept of a tool.

PR #38, opened 2020-01-08
dockstore/dockstore#2959 added terra to partners

emlys:
Also added whitespace between all the logos because they looked squished.
Screen Shot 2020-01-08 at 10 54 53 AM


coverbeck:
@abaumann Could you please approve this PR via a comment? We have our GitHub repo configured so that only members can be reviewers, which is why we're not just adding you as a reviewer.

Thanks.

PR #41, opened 2020-01-10
change version to empty

agduncan94:
Update release version
https://github.com/dockstore/dockstore/issues/3128

Version is being included in the docs title. Set to null so that docs are not versioned.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,9 +24,9 @@ copyright = u'2019, OICR'
 author = u'Andrew Duncan'
 
 # The short X.Y version
-version = u'1.6.0'
+version = u'1.8.0'
 # The full version, including alpha/beta/rc tags
-release = u'1.6.0'
+release = u'1.8.0'
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


denis-yuen:
Actually, I didn't carefully read the ticket. Is the title really the only place this shows up?
If so, maybe we just shouldn't have it and just mark features with what release they are available in. That way, it is harder for docs to be out-of-date or ahead of themselves

agduncan94:
Since we don't do releases for the docs, there is actually nowhere visible to the user where we say 1.8.0 for example. It is only in the page source.

garyluu:
> Ok, then I'd recommend experimenting with just taking it out and seeing if it affects anything other than
> Screenshot from 2020-01-13 10-35-51

Ah, I can't even see that in Chrome.  The title gets truncated.

agduncan94:
@garyluu that is because you have hundreds of windows open :laughing:

coverbeck:
It also shows up if you link to the docs from a Google Doc.

agduncan94:
@denis-yuen It looks like setting the version and release to blank the title looks like 'Welcome to Dockstore Documentation - Dockstore Documentation'. Is this better?

PR #42, opened 2020-01-21
Feature/specify java8 plugins

garyluu:
Additional clarification on the Java 8 version being used to build plugins.

For dockstore/dockstore#3158


docs/advanced-topics/developing-file-provisioning-plugins.rst
```diff
@@ -10,7 +10,7 @@ the
 as a model for plugins where a Java library is not available and the
 plugin needs to call out to an external binary.
 
-This was developed in an environment with Java 8 and Maven 3.3.9.
+This was developed and built in an environment with Java 8 and Maven 3.3.9. Newer versions of Java will not work because of the 1.x.x pf4j plugin used.
```
coverbeck: If I understand correctly and assuming it's the case, I think you should be more explicit that the only the build-time Java needs to be 8, but that it will run on any Java >=8.

PR #43, opened 2020-01-23
Document Launch with AnVIL button

coverbeck:
dockstore/dockstore#3115

Document Launch with AnVIL button. As part of this:

* Now use the same Dockstore screenshot for all the launch with for WDL
* Changed the See Also on the launch pages to only reference other Launch withs,
since there are now a fair number of Launch with's and Azure Batch and AWS Batch seem
less applicable
* Updated copyright
* Fixed build warning


docs/launch-with/anvil-launch-with.rst
```diff
@@ -0,0 +1,61 @@
+AnVIL
+=====
+
+Dockstore integrates with the the Analysis, Visualization, and Informatics Lab-space (AnVIL) platform,
```
denis-yuen: Maybe a link would be helpful
https://www.genome.gov/Funded-Programs-Projects/Computational-Genomics-and-Data-Science-Program/Genomic-Analysis-Visualization-Informatics-Lab-space-AnVIL perhaps

coverbeck: Added link

denis-yuen:
Pressed wrong button trying to re-review

docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # -- Project information -----------------------------------------------------
 
 project = u'Dockstore'
-copyright = u'2019, OICR'
+copyright = u'2020, OICR'
 author = u'Andrew Duncan'
```
agduncan94: We should update the author.

denis-yuen:
Extra thought, there is also https://docs.dockstore.org/en/develop/end-user-topics/language-support.html

docs/conf.py
coverbeck: To whom?

coverbeck: Per Slack discussion, changed it to `OICR, UCSC`

PR #45, opened 2020-01-29
Update links to point at new repo

david4096:
Updates links to point to https://github.com/dockstore/dockstore-cli addresses https://github.com/dockstore/dockstore/issues/2927


denis-yuen:
Looks like some significant breakage (not due to this PR, but does need to get fixed)
Looks like openwdl is just https, some others might require more thought

PR #46, opened 2020-01-29
Add docs support for WDL output file provisioning

wshands:
File provisioning out is supported for Cromwell WDL if you use the Dockstore CLI optional parameter –wdl-output-target which allows you to specify a remote path to provision output files to ex: s3://oicr.temp/testing-launcher/


denis-yuen:
I'm surprised, do the plugins work for WDL provisioning out?

wshands:
Links have been fixed

PR #57, opened 2020-05-05
GitHub app usability review

agduncan94:
Updates the documentation related to GitHub apps and ./dockstore.yml based on feedback from the usability review. Still missing the service dockstore.yml example.

Edit:
Link to live docs https://docs.dockstore.org/en/github-app-usability-review/


docs/faq.rst
```diff
@@ -297,23 +297,18 @@ directory specified prior to a ``launch --local-entry`` command
 
 Why should I migrate my existing workflow to use GitHub Apps and a .dockstore.yml?
 ----------------------------------------------------------------------------------
-Version 1.9.0 has added support for fetching workflows using GitHub Apps. Installing our GitHub App
+Release 1.9.0 of Dockstore has added support for syncing workflows using GitHub Apps. Installing our GitHub App
```
denis-yuen, line 300: Should also say something about it being more efficient from a user perspective.
i.e. we only pull the versions that have changed, so it is faster

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+        testParameterFiles: <String Array>
 
 version
     The version of the .dockstore.yml schema. Currently at 1.2.
 workflows
     An array of workflows. Each element corresponds to a workflow on Dockstore.
-name
+name (optional)
     The optional workflow name that is used to uniquely identify workflows in repositories with multiple workflows.
+    **Each workflow listed must have a unique name.**
```
denis-yuen, line 77: Maybe `Each workflow listed must have a unique (or no) name.`

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
 
    version: 1.2
    workflows:
-      - name: aligner
-        subclass: CWL
-        primaryDescriptorPath: /Dockstore.cwl
-        testParameterFiles:
-            - /test/dockstore.cwl.json
+      - name: <String>
+        subclass: <CWL | WDL | NFL | GALAXY>
```
garyluu, line 67: GALAXY or GXFORMAT2, I don't even know anymore.

agduncan94: Oops forgot to update that. It is GXFORMAT2

coverbeck:
I find this easier to read/review in parallel on docs.dockstore.org; could you please make it available there?

docs/getting-started/dockstore-workflows.rst
```diff
@@ -68,41 +68,57 @@ GitHub Apps, Dockstore can react to changes on GitHub as they are made, keeping
...
-        name: aligner
+        - /test/aligner.cwl.json
 
-If you had our GitHub App installed on a repository ``myorg/alignments`` and then add the above ``.dockstore.yml`` to the ``develop`` branch,
+If you had our GitHub App installed on the repository ``myorg/alignments`` and then add the above ``/.dockstore.yml`` to the **develop** branch,
 the following would occur.
 
-* A workflow with path ``github.com/myorg/alignments/aligner`` will be created on Dockstore
-* The version ``develop`` is added to the workflow ``github.com/myorg/alignments/aligner``
+* A **CWL** workflow with ID ``github.com/myorg/alignments`` will be created on Dockstore
```
coverbeck, line 110: ...with **the** ID..., I think

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
 you add a new branch or release of a workflow on GitHub, Dockstore is notified,
 and Dockstore updates its copy of the workflow. For example, After publishing a new release
 of a workflow on GitHub, a new version of the workflow will be present in
 Dockstore shortly afterwards.
 
-For this to work, a ``/.dockstore.yml`` file is required in each GitHub repository you want
+For this to work, a ``/.dockstore.yml`` file is required in the root directory of each GitHub repository you want
 to associate with a workflow on Dockstore. A template for both workflows and services are shown below,
-along with explanations for each field. For every branch on GitHub that has one of these files, a corresponding entry
+along with explanations for each field. For every branch/release on GitHub that has one of these files, a corresponding entry
```
coverbeck, line 55: Not sure if worth clarifying, but since it's an array, it can be "corresponding entries". But maybe it's worth keeping simple.

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+    **Each workflow listed must have a unique (or no) name.**
 subclass
-    The descriptor language used for the workflow. Supported values include CWL, WDL, and NFL.
+    The descriptor language used for the workflow. Supported values include CWL, WDL, NFL (Nextflow), and GXFORMAT2 (Galaxy).
 primaryDescriptorPath
-    The path to the primary descriptor file in the Git repository
-testParameterFiles
-    An array of paths to test parameter files in the Git repository.
+    The absolute path to the primary descriptor file in the Git repository
+testParameterFiles (optional)
```
coverbeck, line 82: I think this key is required, although it is not required to have any values. We should probably make it optional. If not, we should change this. Can you please log a ticket if you're going to keep the doc like this?

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+      - subclass: CWL
+        primaryDescriptorPath: /Dockstore.cwl
+        testParameterFiles:
+            - /test/dockstore.cwl.json
+
+The above ``.dockstore.yml`` is for a single workflow. Note that the name is not present since it is optional.
+
+Ex. .dockstore.yml with multiple workflows
+
+.. important:: The **name** field is an optional field used when a repository has multiple workflows in it that a user wants to register
```
coverbeck, line 100: Sort of implies that you can't set it if there is only one workflow. Which you probably wouldn't, but you never know.

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+The pushing of a branch with a valid ``/.dockstore.yml`` should trigger Dockstore to convert your workflow into a GitHub App workflow! New versions will be added/updated/deleted
+as if the workflow was originally added using GitHub Apps.  Existing branches will persist, and you can even individually refresh them on the versions tab of the workflow.
+You can convert these existing workflows to use GitHub Apps by adding a valid ``/.dockstore.yml`` to the corresponding branches on GitHub.
+
+Error Handling
+----------------------------------
+Since Dockstore relies on GitHub to tell us when changes have been made on GitHub, there are chances that the message gets lost or delayed.
+Typically, Dockstore reacts within seconds of a change being made on GitHub, however service disruptions can delay this to a few minutes.
+If a message were to get lost, unfortunately you will need to push to GitHub again. Currently, there is no way to tell on Dockstore whether
+a GitHub message was delayed or lost. We recommend waiting a few minutes and then trying to push again. This will hopefully be changed in the near future.
```
coverbeck, line 144: I would say "This will be changed in the future", dropping the "hopefully" and the "near"

docs/faq.rst
```diff
@@ -297,23 +297,19 @@ directory specified prior to a ``launch --local-entry`` command
...
-it can automatically update the corresponding workflows on Dockstore.
+on Dockstore is always automatically synced with GitHub. This means less work for publishers of content on Dockstore,
+and less waiting for content to update. This requires the addition of a ``/.dockstore.yml`` file to your repository.
+This file contains version information such as workflow path, test parameter paths, etc that Dockstore will use to setup
+the version on Dockstore. If Dockstore sees this file in a repository with GitHub App installed, it can automatically
+update the corresponding workflows on Dockstore. You can read more about it at :doc:`Dockstore GitHub Apps <./getting-started/github-apps>`.
 
 How does this change your development flow? Once you have a ``/.dockstore.yml`` file in your master/develop branches on GitHub, it should be
 there anytime you create a new branch. This means that Dockstore will pull in any new branch. If you want to edit version
 information, such as workflow path, you will have to update the ``/.dockstore.yml`` file directly on GitHub. You can no longer do this directly on Dockstore.
```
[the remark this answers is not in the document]
agduncan94: That is correct though isn't it? Since you have to go to GitHub to edit the .dockstore.yml. Or do you mean it implies they must use the website and not the Git command line?

docs/getting-started/dockstore-workflows.rst
```diff
@@ -68,41 +68,57 @@ GitHub Apps, Dockstore can react to changes on GitHub as they are made, keeping
...
+
+   Select which organization to install our GitHub App on
+
+.. figure:: /assets/images/docs/gh-app-reg-3.png
+   :alt: GitHub App Organization Install Page
+
+   Install our GitHub App on either all repositories in an organization or a select few
+
+Once you've installed our GitHub app on a repository or organization, you'll need to add a ``/.dockstore.yml`` file to
+the root directory of a branch of the repository that contains your workflow. This file contains information like
```
Ldcabansay, line 89: Is it always the root directory of the branch repository? Or the same directory as the primary descriptor? Some people put their descriptors in a sub directory for example.

docs/getting-started/github-apps.rst
agduncan94: I think in this case we keep it simple. Below we show an example with one workflow and another with multiple.

docs/getting-started/github-apps.rst
agduncan94: I believe I have tested without the testParameterFiles key and it works.

docs/getting-started/github-apps.rst
agduncan94: What about something like:

> Though the **name** field is an optional field when a ``.dockstore.yml`` has one workflow in it,  it must be used when a ``.dockstore.yml`` has multiple workflows in it. Each entry within a ``.dockstore.yml`` file corresponds to a unique entry on Dockstore.

docs/getting-started/dockstore-workflows.rst
agduncan94: Good question, it must be in the root directory.

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+        testParameterFiles:
+            - /test/globalAligner.cwl.json
+      - name: localAligner
+        subclass: CWL
+        primaryDescriptorPath: /runLocalAligner.cwl
+        testParameterFiles:
+            - /test/localAligner.cwl.json
+
+A common pattern seen on Dockstore is GitHub repositories that store many workflows. The above ``.dockstore.yml``
+has two entries for workflows. Notice that each entry uses a different name. Names are required if you want
```
Ldcabansay, line 121: Out of curiosity is requiring the name for multiple workflows in the same repo a departure from manual register? This is the sam as 'extra' name right?

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
 Service YML File
 +++++++++++++++++
 TODO
 
+Migrating Existing Workflows to use GitHub Apps
+-------------------------------------------------
+If you have already registered GitHub workflows on Dockstore using our old method which required you to refresh, you are still able to use GitHub Apps!
+The migration process is fairly straightforward. First, make sure to install our GitHub App onto the repository for the workflow that you want to migrate.
+Next, create a ``/.dockstore.yml`` file and push it to a branch on your GitHub repo, making sure that the name field matches the existing workflowname.
+If your workflow did not have a workflowname set, simply exclude the field. It is very important that these names match, or else Dockstore won't know which
```
Ldcabansay, line 134: So a no-name has to match with no-name? Building on previous question, what if multiple workflows without extra names were manually registered from the same repo? When you migrate and have to add a name will it cause a duplication?

docs/getting-started/dockstore-workflows.rst
Ldcabansay: Good to know, I'd suggest maybe changing the screen shot to having the primary descriptor in a sub directory and .dockstore.yml in root so its not ambiguous. Or maybe an additional screen shot explaining both cases.

docs/getting-started/github-apps.rst
agduncan94: This is the same as the extra workflow name, which is required if you have multiple workflows from the same repository. One of these workflows could have no name and it would work, as long as two of them didn't have the same name.

docs/getting-started/github-apps.rst
agduncan94: Yes, a no name must name with a no name. This is because no-name is a specific entry on dockstore, whereas having a name (Ex. Foobar) is a different entry but from the same repository.

It is impossible for multiple workflows without extra names to be registered from the same repo. The paths (ex. github.com/org/foobar/name) must be unique, and if two workflows have the same repo but no name set, the paths are the same.

When you migrate, you must match the existing settings. So if you are migration a workflow with no workflow name set, you would have an entry in the dockstore.yml with no name set, and those map to each other. If you had a dockstore.yml with a name set, then the migration would not happen, since they are not referencing the same workflows. Instead a new workflow with that name will be created.

Ldcabansay: Good to know, I've always added the extra-name so never ran into this. 
Do we just handle this as a blank extra-name or a null value? Cause I wonder what happens when the initial workflow was registered without an extra name, but then later you added another workflow with a specified extra-name (which you said is possible with manual register). Will the migration will still work? 
Sorry just trying to be thorough, and the doc language implies that it wouldn't. If it works then I don't think you need to add those specifics to the doc, but just trying to avoid potential bugs.

docs/getting-started/github-apps.rst
```diff
@@ -45,47 +44,117 @@ sync with the linked GitHub repository.
...
+workflow to associate the ``/.dockstore.yml`` with.
+
+The pushing of a branch with a valid ``/.dockstore.yml`` should trigger Dockstore to convert your workflow into a GitHub App workflow! New versions will be added/updated/deleted
+as if the workflow was originally added using GitHub Apps.  Existing branches will persist, and you can even individually refresh them on the versions tab of the workflow.
+You can convert these existing workflows to use GitHub Apps by adding a valid ``/.dockstore.yml`` to the corresponding branches on GitHub.
```
Ldcabansay, line 139: I'd specify "root directory of the corresponding branches..." just to be paranoid.

docs/getting-started/github-apps.rst
agduncan94: I think it is null. With the case you mentioned, they are treated as separate workflows, so the migration will still work.

PR #59, opened 2020-05-12
image checksums don't work for hosted tools

NatalieEO:


agduncan94:
Is this something we want to eventually support? If so I'd create a ticket.

PR #62, opened 2020-06-08
Add table for Cromwell version used

wshands:
Addresses:
dockstore/dockstore#3204
https://ucsc-cgl.atlassian.net/browse/DOCK-1198

### DIFF
--- a/docs/advanced-topics/advanced-features.rst
+++ b/docs/advanced-topics/advanced-features.rst
@@ -311,14 +311,22 @@ primary descriptor is in the root directory of its git repository.
 WDL Launcher Configuration
 --------------------------
 
-By default, WDL tools/workflows will automatically be run with
-`cromwell <https://github.com/broadinstitute/cromwell>`__ 36.
+By default, WDL tools/workflows will automatically be run by the Dockstore CLI
+with the `Cromwell <https://github.com/broadinstitute/cromwell>`__ version listed below.
+
++-------------+-----------------------+
+| CLI version | Cromwell Version used |
++=============+=======================+
+|     1.8.0   |          44           |
++-------------+-----------------------+
+
+
 Additionally, you can override the cromwell version in your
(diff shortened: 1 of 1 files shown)


wshands:
Screen Shot 2020-06-08 at 6 55 53 PM

garyluu:
> Let's pre-emptively add 1.9.0
> 
> > Approved... assuming I don't have to maintain this.
> 
> Hah, we're all maintaining this

Eek, in that case.  I'd rather we add a command to the CLI that either displays the version of cromwell it's currently using or displays the default version.  this is fine for now though.

wshands:
OK but what version of Cromwell will the 1.9.0 CLI (assuming the next version of the CLI is 1.9.0, I guess it could be 1.8.6?) support?

docs/advanced-topics/advanced-features.rst
```diff
@@ -311,14 +311,22 @@ primary descriptor is in the root directory of its git repository.
...
 
-By default, WDL tools/workflows will automatically be run with
-`cromwell <https://github.com/broadinstitute/cromwell>`__ 36.
+By default, WDL tools/workflows will automatically be run by the Dockstore CLI
+with the `Cromwell <https://github.com/broadinstitute/cromwell>`__ version listed below.
+
++-------------+-----------------------+
+| CLI version | Cromwell version used |
++=============+=======================+
+|     1.8.0   |          44           |
```
coverbeck, line 320: Perhaps 1.8 or 1.8.x? Has it changed in the 1.8 timeframe? If not, I lean towards 1.8.

denis-yuen:
> OK but what version of Cromwell will the 1.9.0 CLI (assuming the next version of the CLI is 1.9.0, I guess it could be 1.8.6?) support?

I don't think we have time to update, so I would assume that the CLI we release with the 1.9.0 webservice will be the same version of Cromwell, whatever we call it.

PR #64, opened 2020-06-09
Add Galaxy to language support table

wshands:
Addresses:
dockstore/dockstore#3531
https://ucsc-cgl.atlassian.net/browse/DOCK-1397


wshands:
Screen Shot 2020-06-09 at 10 06 14 AM

garyluu:
I think the Tools Tab is limited support at best.  DAG tab is further along that tools tab.

denis-yuen:
> Looks like it was like this, but shouldn't CLI support for NextFlow be none? And no for the file provisioning plugins?

I'd have to look at the launch-with tab, but I think we displayed launch-from "github" which is a bit of a cheat

PR #66, opened 2020-06-15
dockstore best practices

NatalieEO:
dockstore/dockstore#3334

Viewable here
https://docs.dockstore.org/en/feature-dockstore-best-practices/advanced-topics/best-practices/best-practices-dockstore.html


bethsheets:
There's a typo in this sentence (on instead of one): "Because it will be on the of first things a user will see when looking at your entry, you should make it as detailed as possible."

Looks great otherwise! Thanks!

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+This allows for better organization and collaboration, and also provides a fallback for others if you become inactive on the git repository site.
+
+We also highly recommend that you do NOT host different tools and workflows on a single repo by putting them on different branches.
+Our reasoning is that branches are supposed to be used when working on a new feature or bug fix. That means they should eventually be merged with the repository's main branch.
+Another reason why we are against this practice is because it can lead to a lot of clutter on the Dockstore entries themselves.
+Let's say you have three workflows in a single repo on separate branches and there are many tags/releases related to each one.
+Depending on how you register the workflows on Dockstore, each entry could have the other two branches and every single tag/release
+on its Versions tab. All of this clutter can be overwhelming or confusing for users to see.
+
+Likewise, we don't recommend using even a single branch to hosts multiple entries. A repository should essentially represent one project, but that practice can be a grey area.
```
coverbeck, line 21: Grammatically,  "hosts" should be "host", but I would say "contain"

Regarding this and the previous paragraph, seems like we are taking a hard stance a common pattern. I agree with the previous paragraph more; this one I'm not so sure. The reason people do it this way it to share code.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Improving Your Entries On Dockstore
+-----------------------------------
+
+Making It Easier to Find
+^^^^^^^^^^^^^^^^^^^^^^^^
+There are a few things you can do to make your tool/workflow easier to find on our site.
+
+Let's start with making your entry easier to find on our `search <https://dockstore.org/search>`_ page. One way is to add author and description metadata to your descriptor file.
+Adding an author will make it selectable on the Author facet and a description helps because doing a text search sifts through the text of your descriptor files.
```
coverbeck, line 40: How does this interact with the README.md? Are the contents of the README.md indexed?

Should you cover using the README.md to document things as an alternative?

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+Another tip was already mentioned above; host your repositories on a non-personal organization. Similarly, try to use a non-personal namespace to register your Docker images.
+Doing this will group your tools/workflows together under our ``Tool: Namespace`` and ``Workflow: Organization`` facets.
+
+You should also consider adding labels to your entry since ``Labels`` is another facet on the search page. You can do this by going to `My Tools <https://www.dockstore.org/my-tools>`_ or `My Workflows <https://www.dockstore.org/my-workflows>`_ page.
+On the right hand side, above the tabs, you will see the text``Manage labels``.
+
+.. image:: /assets/images/docs/manage-labels.png
+
+Once you click the text, you'll be able to add or remove labels for your entry. To get ideas for what to use as labels, look at what others are using on the search page.
+Also note that each label cannot have spaces and must be all lowercase. If you want a label to be two words, use a hyphen.
```
coverbeck, line 56: > If you want a label to be two words, use a hyphen.

Nitpick -- using the hyphen doesn't make it two words. Maybe something lke, "if you want to use multiple words, combine them with hyphens" (could be better).

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Another option for making your work more visible is using our :doc:`Organization and Collection <../organizations-and-collections>` pages.
+An Organization is a landing page for your collaborations, institutions, consortiums, companies, etc. Here you can explain the work and goals your group has,
```
coverbeck, line 59: It's your chance to use the word "consortia". :) Both ways are good though, don't change.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+and highlight your tools and workflows by adding them to a Collection.
+
+Making It Easier to Understand
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Once a user has found your entry, they need to understand what it does and how to use it! The most important thing you can do so others understand your work is to provide a thorough description.
+You can do this by filling out the metadata field as explained in the best practices tutorials linked above, or you can write one in the README file.
```
coverbeck, line 65: README.md

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+Your description can be formatted using markdown. Once registered, it will be parsed by Dockstore and made available on the Info tab of an entry.
```
coverbeck, line 66: > Your description can be formatted using markdown

At this point, it's a little vague if you are referring to the description field or README.md. Although both apply, except WDL has a one line limitation, and I don't know about Nextflow.

As I write this, I feel like we should push people to use the README.md, although that's not well thought out and I don't know what others think.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+Because it will be on the of first things a user will see when looking at your entry, you should make it as detailed as possible. Here is a list of items to write about:
+
+* About section.
+
+   * What does your tool or workflow do?
+   * Are you part of a bigger organization? What are some of their goals
```
coverbeck, line 72: Add second question mark

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+and highlight your tools and workflows by adding them to a Collection.
+
+Making It Easier to Understand
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Once a user has found your entry, they need to understand what it does and how to use it! The most important thing you can do so others understand your work is to provide a thorough description.
+You can do this by filling out the metadata field as explained in the best practices tutorials linked above, or you can write one in the README file.
+Your description can be formatted using markdown. Once registered, it will be parsed by Dockstore and made available on the Info tab of an entry.
+Because it will be on the of first things a user will see when looking at your entry, you should make it as detailed as possible. Here is a list of items to write about:
+
+* About section.
```
coverbeck, line 69: You have "Section" below, but "section" here.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+* Citations
+
+   * Does your workflow employ packages that should be cited?
+
+Making It Easier to Use
+^^^^^^^^^^^^^^^^^^^^^^^
+Although it's not always possible, you should provide the input data needed to run the entry. You can do this a few different ways:
+
+* Provide links to the data needed in your description.
+* Have your entry download the input files using a link. You can do this directly in the descriptor files or by putting them in a test parameter file.
```
coverbeck, line 102: I feel like the test parameter JSON should be the number one recommended way

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+  * Link out to other similar entries you think could be useful to others.
+
+* Contact Section
+* Citations
+
+   * Does your workflow employ packages that should be cited?
+
+Making It Easier to Use
+^^^^^^^^^^^^^^^^^^^^^^^
+Although it's not always possible, you should provide the input data needed to run the entry. You can do this a few different ways:
```
coverbeck, line 99: > should provide the input data

Maybe "should provide input data". Or "at least some sample input data".

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+If that is the case, and you want to keep them tied together this way, make sure to give the descriptor files helpful file names.
+It can also help to keep them organized using different, well-named folders for each entry.
+
+One exception to this rule is when you describe the same tool or workflow with multiple descriptor languages. For example, we describe the same bamstats tool
+in our :doc:`Getting Started with CWL <../../getting-started/getting-started-with-cwl>` and :doc:`Getting Started with WDL <../../getting-started/getting-started-with-wdl>` tutorials.
+A single tool on Dockstore can have CWL and WDL descriptor files. Users will be able to see you've provided both and can easily switch between the descriptors in the UI.
+Unfortunately, you cannot have a single workflow entry on Dockstore contain multiple descriptor types. You will have to register them as separate workflows.
+
+
+Improving Your Entries On Dockstore
```
coverbeck, line 32: I think this needs a section on checker workflows.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   * Does your workflow employ packages that should be cited?
+
+Making It Easier to Use
+^^^^^^^^^^^^^^^^^^^^^^^
+Although it's not always possible, you should provide the input data needed to run the entry. You can do this a few different ways:
+
+* Provide links to the data needed in your description.
+* Have your entry download the input files using a link. You can do this directly in the descriptor files or by putting them in a test parameter file.
+* Have the files within the Docker image being used. If you do this, make sure you provide a description of the structure and expected files in the description above.
```
coverbeck, line 103: I don't know that we should be recommending this... Did this use to be there in some form?

1. Depending on data, it can make the image size grow a lot
2. Tends to make it hard-coded for a particular set of data, although it could be overridden.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+If a lot of different workflows or tools work together or are closely related, it can be argued that they're all part of a single project.
+If that is the case, and you want to keep them tied together this way, make sure to give the descriptor files helpful file names.
+It can also help to keep them organized using different, well-named folders for each entry.
+
+One exception to this rule is when you describe the same tool or workflow with multiple descriptor languages. For example, we describe the same bamstats tool
+in our :doc:`Getting Started with CWL <../../getting-started/getting-started-with-cwl>` and :doc:`Getting Started with WDL <../../getting-started/getting-started-with-wdl>` tutorials.
+A single tool on Dockstore can have CWL and WDL descriptor files. Users will be able to see you've provided both and can easily switch between the descriptors in the UI.
+Unfortunately, you cannot have a single workflow entry on Dockstore contain multiple descriptor types. You will have to register them as separate workflows.
```
agduncan94, line 29: Might want to mention that users can use workflowName to register multiple workflows from the same repo.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,107 @@
...
+Another option for making your work more visible is using our :doc:`Organization and Collection <../organizations-and-collections>` pages.
+An Organization is a landing page for your collaborations, institutions, consortiums, companies, etc. Here you can explain the work and goals your group has,
+and highlight your tools and workflows by adding them to a Collection.
+
+Making It Easier to Understand
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Once a user has found your entry, they need to understand what it does and how to use it! The most important thing you can do so others understand your work is to provide a thorough description.
+You can do this by filling out the metadata field as explained in the best practices tutorials linked above, or you can write one in the README file.
+Your description can be formatted using markdown. Once registered, it will be parsed by Dockstore and made available on the Info tab of an entry.
+Because it will be on the of first things a user will see when looking at your entry, you should make it as detailed as possible. Here is a list of items to write about:
```
agduncan94, line 67: 'on' should be one

docs/advanced-topics/best-practices/best-practices-dockstore.rst
NatalieEO: No, it hasn't been mentioned before. I had asked @wshands for some suggestions and he mentioned this as an option.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
NatalieEO: I wrote this section based off memory of conversations (not exactly reliable)... I thought we almost never liked when users had multiple workflows in a repo. If it's just not recommended when they use separate branches, then I can remove this paragraph.

coverbeck: It's caused us some issues server-side, but it seems to be a fairly common pattern. Whether we want to recommend for or against it, I'm not sure.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
NatalieEO: Just checked. I updated a workflow README to have a word that didn't produce results. Once it was refreshed, it came up by searching for that word

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,111 @@
...
+For more detailed information on these metadata fields, check out the following info for each language:
+
+- :doc:`WDL Best Practices <./wdl-best-practices>`
+- :doc:`CWL Best Practices <./best-practices>`
+- :doc:`Nextflow Best Practices <./nfl-best-practices>` and `Nextflow manifest documentation <https://www.nextflow.io/docs/latest/config.html#scope-manifest>`_
+
+You can also provide a description by writing a README.md file instead. If you do not provide description metadata in your descriptor, then we will try to pull the README.md file as a fallback.
+
+Another tip was already mentioned above; host your repositories on a non-personal organization. Similarly, try to use a non-personal namespace to register your Docker images.
+Doing this will group your tools/workflows together under our ``Tool: Namespace`` and ``Workflow: Organization`` facets.
```
denis-yuen, line 50: Drawing on recent events, it also has the advantage of allowing you to add multiple developers that can manage your content on and off Dockstore in case you get hit by a bus

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,111 @@
...
+   * Are you part of a bigger organization? What are some of their goals?
+
+* How to Use Section.
+
+   * What are the system requirements? Minimum and recommended
+   * Describe the input and output files
+
+      * What are their names?
+      * What data do they contain?
+      * What is the format?
```
denis-yuen, line 86: Can point out that in CWL you can include this info in the descriptor which is even better because search understands it
https://www.commonwl.org/user_guide/16-file-formats/index.html

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,111 @@
...
+   * Can you provide time and/or cloud cost estimations for running your tool/workflow with a given input?
+   * If available, link to tutorials using your entry.
+   * If available, link to a sample or complete dataset to use.
+
+* Related To Section
+
+   * Does your tool/workflow work together with other entries? If so, describe how they can be used together and provide links.
+   * Link out to other similar entries you think could be useful to others.
+
+* Contact Section
```
denis-yuen, line 97: Contact info can also be from descriptors

docs/advanced-topics/best-practices/best-practices-dockstore.rst
denis-yuen: I'd say soft-recommend against. 
I would say that 
a) putting workflows in different repos gives shorter names since you don't need an extra name to distinguish
b) you don't get weird extra releases when you release/tag one workflow but don't really mean to refresh the other one

But it seems like people like multiple workflows in one repo to share code, so :shrug:

docs/advanced-topics/best-practices/best-practices-dockstore.rst
```diff
@@ -0,0 +1,114 @@
...
+given some input, produces the expected output on a platform different from the one where you are developing. Read our :doc:`Checker Workflows <../checker-workflows>` tutorial to learn more.
+
+Another option for making your work more visible is using our :doc:`Organization and Collection <../organizations-and-collections>` pages.
+An Organization is a landing page for your collaborations, institutions, consortiums, companies, etc. Here you can explain the work and goals your group has,
+and highlight your tools and workflows by adding them to a Collection.
+
+Making It Easier to Understand
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Once a user has found your entry, they need to understand what it does and how to use it! The most important thing you can do so others understand your work is to provide a thorough description.
+You can do this by filling out the metadata field as explained in the best practices tutorials linked above, or you can write one in the README.md file.
```
coverbeck, line 73: If I have this right(@garyluu, do I ?), I think you should explicitly say that the README.md is only used if there is no description.

Otherwise it may not be clear when the README.md is used or not.

garyluu: "If no description from the descriptor file, fall back to README".  I don't think README is related to the other metadata.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
agduncan94: I agree, but maybe as another ticket.

docs/advanced-topics/best-practices/best-practices-dockstore.rst
wshands: Including reference genome files, etc. in the Docker image is what is being done for the HuBMAP ATAC-seq pipeline; it dramatically increases the size of the Docker image but reduces the number of inputs needed. This was done to make it very easy for the team that runs the workflows to work with the pipeline. I don't know how common it is to use this technique.

PR #72, opened 2020-07-06
1.9.0 news

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1489
Viewable here: https://docs.dockstore.org/en/feature-1.9.0-news/news/2020-07-06-Dockstore-1.9.0.html

### DIFF
--- a/docs/news/2020-07-06-Dockstore-1.9.0.rst
+++ b/docs/news/2020-07-06-Dockstore-1.9.0.rst
@@ -4,19 +4,21 @@ Dockstore 1.9.0
 Highlighted new features include:
 ---------------------------------
 
--  Language plugin support, allowing for additional support of Galaxy workflows as a preview feature
+-  Preview feature for Galaxy workflow support
 -  Performance improvements for a large variety of endpoints
 -  GitHub app support in preview to allow for automated update of workflow content
 -  Implementation of the `TRS v2 final standard <https://github.com/ga4gh/tool-registry-service-schemas/releases/tag/2.0.0>`_
 
    -  TRS v2 beta standard support is deprecated, but still present
 
--  Capture of file and docker image checksums from GitHub, quay.io, Gitlab, and Docker Hub on workflow snapshot to support immutable workflows (see breakdown for details here_) support for linking ORCID profiles user profiles
+-  Capture of file and docker image checksums from GitHub, quay.io, Gitlab, and Docker Hub on workflow snapshot to support immutable workflows (see breakdown for details here_)
+-  Support for linking ORCID profiles user profiles
 
    -  Currently displayed in organization and stargazer views
 
 -  A large number of usability improvements and fixes to the user interface website
 -  Partial API migration from swagger 2.0 to openapi 3.0 for the Dockstore API
+-  Improved language plugin support so it's easier for Dockstore to support additional languages (like Galaxy)
 -  A large variety of security updates and bug fixes
 
 As always, see our full list of changes on `GitHub`_
@@ -33,7 +35,7 @@ Minor
 ~~~~~
 
 - If upgrading the Dockstore CLI, our Dockstore script has changed and should be downloaded anew from the onboarding wizard
-- CLI website management features (refresh) will be broken until CLI version is upgraded, workflow launching should remain
+- The CLI refresh command will be broken until you update to CLI version 1.9.
 
 .. _GitHub: https://github.com/dockstore/dockstore/milestone/31
 .. _here: https://docs.dockstore.org/en/develop/advanced-topics/checksum-support.html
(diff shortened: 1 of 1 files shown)


docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,39 @@
+Dockstore 1.9.0
+===============
+
+Highlighted new features include:
+---------------------------------
+
+-  Language plugin support, allowing for additional support of Galaxy workflows as a preview feature
```
coverbeck, line 7: Language plugin support is an implementation detail. Even though it's important, it's not a user-facing feature; support for Galaxy is.

tldr; I would just mention Galaxy here, and have a section further below on implementation details.

I guess it depends who we think the audience is.

docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,39 @@
...
+-  Performance improvements for a large variety of endpoints
+-  GitHub app support in preview to allow for automated update of workflow content
+-  Implementation of the `TRS v2 final standard <https://github.com/ga4gh/tool-registry-service-schemas/releases/tag/2.0.0>`_
+
+   -  TRS v2 beta standard support is deprecated, but still present
+
+-  Capture of file and docker image checksums from GitHub, quay.io, Gitlab, and Docker Hub on workflow snapshot to support immutable workflows (see breakdown for details here_) support for linking ORCID profiles user profiles
```
coverbeck, line 14: `support for linking ORCID...` seems like a run-on; should be a separate item?

docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Major
+~~~~~
+
+None intended
+
+Minor
+~~~~~
+
+- If upgrading the Dockstore CLI, our Dockstore script has changed and should be downloaded anew from the onboarding wizard
```
coverbeck, line 35: Is this necessary? I thought we handled it.

docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,39 @@
...
+- CLI website management features (refresh) will be broken until CLI version is upgraded, workflow launching should remain
```
coverbeck, line 36: I feel this is phrased slightly oddly, talking about website management features, and not sure what "workflow launching should remain" means (it will continue to work?). I would just say something like:

The CLI `refresh` command only works in 1.9 (needs work too).

docs/news/2020-07-06-Dockstore-1.9.0.rst
NatalieEO: Not sure...I copied all this over from our GitHub changelog. @wshands @denis-yuen

denis-yuen: Hmmm, taking a closer look. I think if they have a script from 1.8.0 and above, they don't need a new script. 
If they have a script from 1.7.0 and below, they do. 
Unfortunately, our "upgrade" command replaces the jar and not the script. So they could have a 1.8.0 jar with a 1.7.0 script for example. Safest is to download a new script I think

docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,41 @@
...
+
+-  Preview feature for Galaxy workflow support
+-  Performance improvements for a large variety of endpoints
+-  GitHub app support in preview to allow for automated update of workflow content
+-  Implementation of the `TRS v2 final standard <https://github.com/ga4gh/tool-registry-service-schemas/releases/tag/2.0.0>`_
+
+   -  TRS v2 beta standard support is deprecated, but still present
+
+-  Capture of file and docker image checksums from GitHub, quay.io, Gitlab, and Docker Hub on workflow snapshot to support immutable workflows (see breakdown for details here_)
+-  Support for linking ORCID profiles user profiles
```
coverbeck, line 15: "profiles user profiles" -- don't think you want "profiles" in there twice.

Or do you mean "linking ORCID profiles **to** user profiles"? If so, I would say "linking ORCID account to your Dockstore user profile" or something like that.

docs/news/2020-07-06-Dockstore-1.9.0.rst
```diff
@@ -0,0 +1,41 @@
...
+
+   -  Currently displayed in organization and stargazer views
+
+-  A large number of usability improvements and fixes to the user interface website
```
coverbeck, line 19: "...to the user interface website" doesn't sound right; maybe:

* ...to the website's user interface
* ...to the website
* ...to the user interface

PR #76, opened 2020-08-10
Feature/3484/explain workflow name

agduncan94:
https://github.com/dockstore/dockstore/issues/3484

Better explain workflow name in our initial Workflow Registration tutorial. Existing GitHub apps docs did a great job of explaining workflow names, so I mostly left those untouched.

I've attached a screenshot of the main changes since I don't want to create a new branch on RTD (might forget to delete). This is on the Workflow Registration tutorial and is the first section in "Register Your Workflow in Dockstore".
Screen Shot 2020-08-10 at 3 59 00 PM

### DIFF
--- a/docs/getting-started/dockstore-workflows.rst
+++ b/docs/getting-started/dockstore-workflows.rst
@@ -167,7 +167,7 @@ removed from Dockstore.
 
 .. tip:: Since the workflows field is an array, this file supports multiple workflows on Dockstore stemming from
    the same repository on GitHub. This is useful if you store a lot of your workflows in the same GitHub
-   repository. This is achieved by setting a different value for the name field for each entry (corresponding to the workflow name).
+   repository. This is achieved by setting a different value for the name field for each entry (corresponding to the workflow name of the entry).
 
 .. important:: The GitHub user who first adds a workflow onto Dockstore must correspond to a user on Dockstore.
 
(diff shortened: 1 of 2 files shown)


docs/getting-started/dockstore-workflows.rst
```diff
@@ -127,9 +167,9 @@ removed from Dockstore.
 
 .. tip:: Since the workflows field is an array, this file supports multiple workflows on Dockstore stemming from
    the same repository on GitHub. This is useful if you store a lot of your workflows in the same GitHub
-   repository. This is achieved setting a different value for the name field for each entry.
+   repository. This is achieved setting a different value for the name field for (Corresponds to the workflow name).
```
coverbeck, line 170: 1. Was like this, but should be "...achieved **by** setting a different value..."
2. "(Corresponds to the workflow name)" -- is this complete? Sentence doesn't quite read right, and why is is a capital C in "Corresponds"? Is this a RST link or something that I shouldn't be reading literally?

agduncan94: I was just trying to emphasize that the name field in the dockstore.yml corresponds to the workflow name property of the workflow.

Looks like I accidentally deleted "each entry"

agduncan94: > This is achieved by setting a different value for the name field for each entry (corresponds to the workflow name).

PR #78, opened 2020-08-28
update organization roles

NatalieEO:
#3742

### DIFF
--- a/docs/advanced-topics/organizations-and-collections.rst
+++ b/docs/advanced-topics/organizations-and-collections.rst
@@ -90,18 +90,18 @@ can request a re-review.
 Organization membership
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Anyone can see an approved organization, though only members and
-maintainers of the organization can perform actions on the
-organizations. This includes creating and adding to collections,
-updating metadata, and adding new members.
+Anyone can see an approved organization, but only admins and
+maintainers of the organization can perform actions on it.
+This includes creating and adding to collections, updating metadata, and adding new members.
 
-There are two types of roles available:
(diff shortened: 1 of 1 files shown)


docs/advanced-topics/organizations-and-collections.rst
```diff
@@ -90,18 +90,18 @@ can request a re-review.
...
-Anyone can see an approved organization, though only members and
-maintainers of the organization can perform actions on the
-organizations. This includes creating and adding to collections,
-updating metadata, and adding new members.
+Anyone can see an approved organization, but only admins and
+maintainers of the organization can perform actions on it.
+This includes creating and adding to collections, updating metadata, and adding new members.
 
-There are two types of roles available:
+As of 1.9.2, there are three types of roles available:
```
garyluu, line 97: Doesn't our documentation versioning take care of this so we don't have to mention the version?

agduncan94: It can take care of versioning, but we don't yet do it that way. But maybe we should do it for releases.

coverbeck: I think there is a story or issue to start tagging our doc releases, isn't there?

denis-yuen:
> Approved, but I think a table showing the differences between them would work too.

 For better or for worse, there is only the one dockstore deployment for now (that is not private). 
So I think while a table would be nice, I think we can skip it for now (it's not like the user has a choice of server versions)

docs/advanced-topics/organizations-and-collections.rst
NatalieEO: Hm...I couldn't find one. I know we've discussed it before.

PR #82, opened 2020-10-08
removed unused $schema declarations

mr-c:


coverbeck (approved):
I don't know CWL well enough to comment to say if this is correct, but it looks like it. :)

denis-yuen (approved):

PR #85, opened 2020-10-29
Fixed local links so that linkcheck no longer throws a errors

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1621

Links can be represented with different roles, [sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html), the command `make linkcheck` specifically expects external links, which is why all our local links were flagged as broken. Here are some approaches that can be used for links:
1. external links:
<code>\`text \<url\>\`</code>
2. link to a local document:
<code>:doc:\`text \</path/to/rst/file\>\`</code>
3. link to specific place in a local document:
   3a. Create a label in the document you want to link to with the following: <code>.. _<b>label-name</b>:</code> ( _ before label name is required)
   3b. link to the label with: <code>:ref:\`text \<label-name\>\`</code>

### DIFF
--- a/docs/advanced-topics/advanced-features.rst
+++ b/docs/advanced-topics/advanced-features.rst
@@ -1,6 +1,8 @@
 Advanced CLI Features
 =====================
 
+.. _file-provisioning:
+
 File Provisioning
 -----------------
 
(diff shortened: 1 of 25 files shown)


garyluu (approved):
From what I gather (since we don't used a tagged sphinx version), sphinx 3.2.0 added local link check but it has problems that won't be fixed until 3.3.0.  This is fine for now, but hopefully the :ref: syntax won't be needed

denis-yuen (approved):

PR #86, opened 2020-11-03
Clarify Nextflow Descriptor for GitHub Apps

nolwarre:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-1802

Differed from the examples and was not clear on which file to use for Nextflow descriptor.

### DIFF
--- a/docs/getting-started/github-apps/github-apps.rst
+++ b/docs/getting-started/github-apps/github-apps.rst
@@ -80,9 +80,14 @@ name (optional)
     this field is required to uniquely identify workflows in the repository.
     **Each workflow listed must have a unique (or no) name.**
 subclass
-    The descriptor language used for the workflow. Supported values include CWL, WDL, NFL (Nextflow), and GXFORMAT2 (Galaxy). This cannot be changed once the workflow is registered.
+    The descriptor language used for the workflow. Supported values include CWL, WDL, NFL (Nextflow), and GALAXY (Galaxy). This cannot be changed once the workflow is registered.
 primaryDescriptorPath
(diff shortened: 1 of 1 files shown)


docs/getting-started/github-apps/github-apps.rst
```diff
@@ -82,7 +82,7 @@ name (optional)
 subclass
     The descriptor language used for the workflow. Supported values include CWL, WDL, NFL (Nextflow), and GXFORMAT2 (Galaxy). This cannot be changed once the workflow is registered.
 primaryDescriptorPath
-    The absolute path to the primary descriptor file in the Git repository.
+    The absolute path to the primary descriptor file in the Git repository. For Nextflow, the primary descriptor is the .config file.
```
coverbeck, line 85: Isn't it typically `nextflow.config`? If so, we should say that.

This feels a little out of place to me. I feel like like we should mention all languages or none. Maybe have sub-bullets? For Galaxy, the primary descriptor is a .ga file; for WDL the primary descriptor file is a .wdl file, etc.

But if I'm the only who feels this way, I'm OK with.

coverbeck (commented):
Not part of your changes, but because I happened to notice it, can you please change `GXFORMAT2` to `GALAXY` on line 70?

wshands (approved):

docs/getting-started/github-apps/github-apps.rst
wshands: 👍

wshands:
Yes please fix line 70 as Charles mentioned

denis-yuen (approved):

coverbeck (approved):

PR #97, opened 2021-03-04
Update language support table

coverbeck:
Update language support table

dockstore/dockstore#3269
dockstore/dockstore#4066
dockstore/dockstore#3900

* Added Cavatica, Galaxy, and BD Cat pages and updated table
* Reworked table a bit, so it doesn't say `(workflows only)` for every platform.
* Fixed all make warnings, e.g., ORCID not in TOC, bad internal links
*  New images
* Removed limitation that Terra and its skins don't support file-based imports. Although only GitHub based workflows don't have that limitation, that's almost all of them, so I chose not to go into that level of detail (but can change).
* Empty [line bug](dockstore/dockstore#3900) on home page
 
Live here: https://docs.dockstore.org/en/feature-4066-language-support/


docs/launch-with/anvil-launch-with.rst
```diff
@@ -45,12 +45,7 @@ Limitations
 
 1. While we support launching of WDL workflows, tools as listed in
    Dockstore are currently not supported.
-2. AnVIL does not currently support file-path based imports. Importing a
-   workflow with file-based imports will result in error. See the
-   :ref:`converting file-based imports
-   doc <converting-file-path-based-imports-to-public-http-s-based-imports-for-wdl>`
-   for more info.
-3. Only the WDL language is supported.
+2. Only the WDL language is supported.
```
bethsheets, line 49: Do you want to add that parameter files (json) are not included in the launch with? 

3. Example parameter files (json) are not included in the launch-with service. After a workflow launch is complete, users can download parameter files from Dockstore and upload them into their Terra workspace that contains the workflow.

docs/launch-with/bdcat-launch-with.rst
```diff
@@ -0,0 +1,43 @@
...
+-----------------------------------------
+.. figure:: /assets/images/docs/sevenbridges/sb_from_dockstore.png
+   :alt: CWL workflow
+
+   CWL workflow
+
+Limitations
+-----------
+
+1. Only workflows can be exported; Dockstore Tools are not supported.
```
bethsheets, line 35: Even for CWL tools?

docs/launch-with/bdcat-launch-with.rst
```diff
@@ -0,0 +1,43 @@
...
+2. For CWL workflows, NHLBI BioData Catalyst does not currently 
+   support http(s) based imports. Dockstore disables the Launch
+   with NHLBI BioData Catalyst button if the selected version has
+   any http(s) imports.
+
```
bethsheets, line 40: Same comment from AnVIL above about adding info about parameter files.

bethsheets (approved):
All of your changes look good to me (thanks for catching some bad links in best practices!). I did have a few suggestions that are outside of this PR.

docs/launch-with/bdcat-launch-with.rst
coverbeck: > Even for CWL tools?

Correct, we do not have the button for CWL tools. I think it actually could/would work with SB platforms, but we don't have it there, and we would need to try it out.

docs/launch-with/anvil-launch-with.rst
coverbeck: > Do you want to add that parameter files (json) are not included in the launch with?
> 
>     1. Example parameter files (json) are not included in the launch-with service. After a workflow launch is complete, users can download parameter files from Dockstore and upload them into their Terra workspace that contains the workflow.

Hmm, I'll see what others think... If we add it, I don't think it should be a limitation, but I could see it being useful in the general overview of how to import.

NatalieEO: Could be good to add as a tip or something like that on the page

NatalieEO (approved):

garyluu (approved):

docs/end-user-topics/language-support.rst
```diff
@@ -13,25 +13,29 @@ on the Dockstore site and in the Dockstore command-line utility.
...
 
-[4] Does not support file-path or http(s) based imports. See :ref:`DNAstack
+[5] Does not support file-path or http(s) based imports. See :ref:`DNAstack
 Limitations <dnastack-limitations>` for limitations.
 
-[5] All verified Dockstore WDL tools/workflows were tested successfully. However, we anticipate that more testing is needed for WDL workflows that use language features not contained within that dataset.
+[6] Does not support file-path based imports. See :ref:`Terra
+Limitations <terra-limitations>` for limitations.
+
+[7] There is no launch with Galaxy button in Dockstore, but Dockstore workflows can be searched and imported from within Galaxy.
```
denis-yuen, line 33: Nice addition

Sorry to add new work, but that this cell is a conditional yes but is yellow while all the other "nos" are yellow too makes me think either
1) this cell should be a new colour
2) "no" should be grey and this should be yellow

denis-yuen (commented):
> Removed limitation that Terra and its skins don't support file-based imports. Although only GitHub based workflows don't have that limitation, that's almost all of them, so I chose not to go into that level of detail (but can change).

I feel like this is going to shoot us in the foot since even I refer to these pages a lot to see what is supported and what is not. 
Thoughts:
1) Code comment?
2) Leave in on the Terra specific page as fine text?

docs/end-user-topics/language-support.rst
coverbeck: Pushed changes with colors, please take a look.

docs/launch-with/anvil-launch-with.rst
coverbeck: Added tip

denis-yuen (approved):

PR #99, opened 2021-03-19
add .dockstore.yml publish documentation

GFJHogue:
Main ticket: https://github.com/dockstore/dockstore/issues/4100
Relevant feature PR: https://github.com/dockstore/dockstore/pull/4085

Adding documentation for setting workflow/service publish state via `.dockstore.yml`.

### DIFF
--- a/docs/assets/templates/.dockstore.yml
+++ b/docs/assets/templates/.dockstore.yml
@@ -9,6 +9,11 @@ service:
   author:
   description:
 
+  # `publish` is a service-wide setting and will affect ALL branches/tags; only set this as needed in a main branch.
+  # `publish` may be set to true to publish an unpublished service, or false to unpublish a published service.
+  # Omitting the publish setting leaves the publish-state unchanged (recommended for all non-primary branches).
+  publish:
+
   # These are files the Dockstore will index. They will be directly downloadable from Dockstore. Wildcards are not supported.
   files:
 
--- a/docs/getting-started/github-apps/github-apps.rst
+++ b/docs/getting-started/github-apps/github-apps.rst
@@ -68,6 +68,7 @@ For a workflow, the ``/.dockstore.yml`` has the following general structure
    workflows:
       - name: <String>
         subclass: <CWL | WDL | NFL | GALAXY>
+        publish: <Boolean>
         primaryDescriptorPath: <String>
         testParameterFiles: <String Array>
         filters:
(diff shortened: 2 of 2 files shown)


coverbeck (approved):
Looks good, but should we merge it into develop, which is our public view? People might start using it and expecting it to work -- apparently NF-Core already has.

Options:

1. Create a 1.11 branch and merge it there, then merge 1.11 when we release 1.11
2. Label this as a 1.11 feature inline in the doc (looks like that it will be tricky in this case).

GFJHogue:
> Looks good, but should we merge it into develop, which is our public view?
> ...
> 1. Create a 1.11 branch and merge it there, then merge 1.11 when we release 1.11

I agree, but this feels like we'd be treating `develop` like how we treat `master` everywhere else.

Why is `develop` the public view in the first place for dockstore-documentation?

coverbeck:
> Why is `develop` the public view in the first place for dockstore-documentation?

I was wondering that myself as I typed in my initial comment. :) I think it was set up that way in the beginning and we never revisited it.

Seems like should follow the more standard Hubflow model. But now we have links to develop from the outside, so we might be stuck with it, unless readthedocs lets you do redirects.

GFJHogue:
> I was wondering that myself as I typed in my initial comment. :) I think it was set up that way in the beginning and we never revisited it.
> Seems like should follow the more standard Hubflow model. But now we have links to develop from the outside, so we might be stuck with it, unless readthedocs lets you do redirects.

That makes sense. I'll go ahead and make a 1.11 branch for this PR, which we can merge in on release.

denis-yuen:
> Seems like should follow the more standard Hubflow model. But now we have links to develop from the outside, so we might be stuck with it, unless readthedocs lets you do redirects.

Did you have particular links to develop in mind? I wonder if they are under our control/influence

coverbeck:
> Did you have particular links to develop in mind? I wonder if they are under our control/influence

Hmm, I may have assumed wrong. I was thinking of some of @bethsheets documents, although she brought one into our docs, so that might be moot. So it might be safe to change after all.

docs/assets/templates/.dockstore.yml
```diff
@@ -9,6 +9,11 @@ service:
   author:
   description:
 
+  # `publish` is a service-wide setting and will affect ALL branches/tags; only set this as needed in a main branch.
```
denis-yuen, line 12: maybe `dockstore-wide`
service is like tool or workflow for us

denis-yuen (approved):
minor comment change

docs/assets/templates/.dockstore.yml
GFJHogue: I indeed mean service-wide here, just as I say workflow-wide elsewhere in this PR.

Maybe there's a better way to word it?

denis-yuen: Oh no, I just missed that this is legit the one for services, carry on

denis-yuen (approved):

Richard-Hansen (approved):

NatalieEO (approved):

garyluu (approved):

PR #105, opened 2021-05-19
Add help on checksum verification

david4096:
We added notes on checking the presence of checksums but not verifying them. This improves the status of that slightly and mentions the CLI feature verifying descriptor checksums.

https://ucsc-cgl.atlassian.net/browse/SEAB-1510

### DIFF
--- a/docs/advanced-topics/checksum-support.rst
+++ b/docs/advanced-topics/checksum-support.rst
@@ -68,7 +68,7 @@ use Docker images referred to by a tagged version or digest. The verification fe
 
 For more information on Docker registry retention policies see posts from `Docker <https://www.docker.com/blog/scaling-dockers-business-to-serve-millions-more-developers-storage/`_,
 `AWS <https://aws.amazon.com/blogs/compute/clean-up-your-container-images-with-amazon-ecr-lifecycle-policies/>`_,
-or `Azure <https://docs.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy>`_ documentation.
+or `Azure <https://docs.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy>`_.
 
 Tools
 -----
--- a/docs/advanced-topics/checksum-support.rst
+++ b/docs/advanced-topics/checksum-support.rst
@@ -61,10 +61,14 @@ Docker CLI client.
 When the download has completed a Digest is provided in the terminal output. This should match the checksum provided
 by the Dockstore API.
 
-Verifying the image checksum can give you better guarantees the image has not changed since it was published to Dockstore.
+Verifying the image checksum can give you better guarantees the image has not changed since the workflow was published to Dockstore.
 However, in some cases the image checksum may diverge, for example, if the image was defined in a git branch that has since
-been updated. For best results use Docker images referred to by a tagged version or digest. The verification features
-available may vary between execution engines.
+been updated. For best results, and to avoid your Docker image being deleted because of a registry's retention policy,
+use Docker images referred to by a tagged version or digest. The verification features available may vary between execution engines.
+
+For more information on Docker registry retention policies see posts from `Docker <https://www.docker.com/blog/scaling-dockers-business-to-serve-millions-more-developers-storage/`_,
+`AWS <https://aws.amazon.com/blogs/compute/clean-up-your-container-images-with-amazon-ecr-lifecycle-policies/>`_,
+or `Azure <https://docs.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy>`_ documentation.
 
 Tools
 -----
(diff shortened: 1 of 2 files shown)


docs/advanced-topics/checksum-support.rst
```diff
@@ -19,6 +19,19 @@ More specifically, the endpoints that contain checksums for files are as follows
 
 The id parameter used in the endpoints above can be found on an entry's public page; underneath the Info tab, look for the bolded words **TRS**.
 
+After gathering the checksum using the above method you can verify a descriptor's checksum using the sha1sum terminal application.
```
coverbeck, line 22: On a Mac, by default, it looks like it's `shasum`. I ran your commands below, and it seems to work with it.

```
curl -s https://dockstore.org/api/ga4gh/trs/v2/tools/$trsid/versions/$version/PLAIN-WDL/descriptor | shasum
c8ea6c657fd2e2ca9743567331e0947aa3eed33a  -
```

Does the trailing dash matter? I guess this is a visual inspection anyway.

But in any case, suggest mentioning shasum and Mac.

docs/advanced-topics/checksum-support.rst
```diff
@@ -37,6 +50,20 @@ Descriptions for the two endpoints of note are as follows:
...
+Docker CLI client.
+
+::
+
+    docker pull quay.io/briandoconnor/dockstore-tool-md5sum:1.0.4
+
+When the download has completed a Digest is provided in the terminal output. This should match the checksum provided
+by the Dockstore API.
+
+Depending on your execution engine, the image may be downloaded using another technique and may not have its checksum
```
coverbeck, line 64: This sounds ominous and I don't quite get it. Are you saying the image can have a different digest based on how it's downloaded?

docs/advanced-topics/checksum-support.rst
david4096: The trailing dash is from the shasum application and it just prints the filename. Since we are piping stdout it just puts a dash. shasum and sha1sum do the same thing on mac and linux so I'll just use shasum. shasum defaults to sha1

docs/advanced-topics/checksum-support.rst
david4096: It could be made more clear. I'll take another try. I'm trying to say "this can perform some verification but its not the same as the execution engine verifying the digest". Since we don't have control over them it's technically possible cromwell could get a MITM attack and this verification wouldn't have proved anything. They don't check to see if the digest on dockstore is the same as what is on Dockerhub/quay and so especially in the case of a branch its possible they could diverge.

docs/advanced-topics/checksum-support.rst
```diff
@@ -37,6 +50,22 @@ Descriptions for the two endpoints of note are as follows:
...
+
+::
+
+    docker pull quay.io/briandoconnor/dockstore-tool-md5sum:1.0.4
+
+When the download has completed a Digest is provided in the terminal output. This should match the checksum provided
+by the Dockstore API.
+
+Verifying the image checksum can give you better guarantees the image has not changed since it was published to Dockstore.
+However, in some cases the image checksum may diverge, for example, if the image was defined in a git branch that has since
```
denis-yuen, line 65: Can also mention/link that if an image is defined off of a branch, Docker Hub is more likely to delete it when they do cleanup. A digest corresponding to a tagged version is probably better for the longevity

docs/advanced-topics/checksum-support.rst
```diff
@@ -37,6 +50,22 @@ Descriptions for the two endpoints of note are as follows:
...
+Docker CLI client.
+
+::
+
+    docker pull quay.io/briandoconnor/dockstore-tool-md5sum:1.0.4
+
+When the download has completed a Digest is provided in the terminal output. This should match the checksum provided
+by the Dockstore API.
+
+Verifying the image checksum can give you better guarantees the image has not changed since it was published to Dockstore.
```
coverbeck, line 64: "...image has not changed since it was published to Dockstore... " -- nit-picky and optional, but I interpret "it" as the image, but I think you mean since the workflow was published to Dockstore.

Or if you do mean the image, it's not published to Dockstore, it's published to DockerHub, Quay, etc.

coverbeck (approved):

NatalieEO (approved):

Richard-Hansen (approved):
Thanks for making these improvements and adding the examples, LGTM.

docs/advanced-topics/checksum-support.rst
david4096: Will clarify that I meant workflow

david4096:
Had to update the conf.py https://github.com/sphinx-doc/sphinx/issues/7747

denis-yuen (approved):

PR #114, opened 2021-06-10
Feature/2990/update UI screenshots

Richard-Hansen:
Scanned through images in the `docs/assets/images/docs` directory and replaced all those that seemed old. One image was deleted, since it wasn't used, the rest were replaced as needed.


denis-yuen:
Ditto, shouldn't this be to https://github.com/dockstore/dockstore-documentation/tree/release/1.11
@garyluu for clarification
Seems like prod is still using develop at this point

garyluu:
> Ditto, shouldn't this be to https://github.com/dockstore/dockstore-documentation/tree/release/1.11
> @garyluu for clarification
> Seems like prod is still using develop at this point

yup, probably release branch

coverbeck (approved):
Pending correct target branch

denis-yuen (approved):

PR #121, opened 2021-07-02
Feature/seab 1977 new logo

bvizzier:
Update logo per conversation with Kim.

Also, update config.py to remove deprecated function call warning.


coverbeck:
Build here: https://docs.dockstore.org/en/feature-seab-1977-newlogo/

docs/assets/images/Dockstore-Documentation-horizontal-white.svg
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
```
coverbeck, line 1: It doesn't look like you're using this file. If you aren't, can you please delete it.

coverbeck (approved):
I'm not entirely wild about the black background "island". But sounds like you discussed with @kcullion and I'm going to defer.

bvizzier:
The reason for the decision is in SEAB-1977.  Basically I find the Dockstore logo color on the ReadTheDocks blue to be clashing.  The dark gray of the ToC bar was the balance to me.  Kim agreed.

I committed the SVG file since it was the source for the generate jpeg that is actually being used.  I checked it in so it would be easily accessible if we wanted to change the jpeg.  I can delete it but wanted to check first.

Ben

coverbeck:
> I committed the SVG file since it was the source for the generate jpeg that is actually being used. I checked it in so it would be easily accessible if we wanted to change the jpeg. I can delete it but wanted to check first.

Ah, I see. It does make sense to keep it and we've run into issues in the past where we couldn't find the original.

It may be a little confusing for somebody coming in fresh and wondering why there are two images with the same name. And we should probably have a central repository where we keep all our artwork -- the UI has lots of graphics (maybe we already do). But that's beyond the scope of this ticket, so  just leave it unless somebody else has a better idea.

bvizzier:
I like the idea of a repo, a section on one of the wikis, or a Google Drive directory specifically for graphics and the sources.

It could also be used for screenshots that are commonly used in presentations and/or documentation.

garyluu (commented):
Is this intended to be for 1.11? if so, then it needs to make its way to the master branch

bvizzier:
> Is this intended to be for 1.11? if so, then it needs to make its way to the master branch

It is not required for 1.11.

denis-yuen (approved):

denis-yuen:
@bvizzier is this ok to merge into develop?

bvizzier:
Yes. Merge away.

Ben

PR #125, opened 2021-08-02
Add information on setting Cromwell's concurrent-job-limit for local testing

aofarrel:
Pretty much any substantial scattered workflow will cause issues if this is not set.

The information is adapted from [this document](https://github.com/DataBiosphere/analysis_pipeline_WDL/blob/main/_documentation_/for%20users/running-locally.md), which probably doesn't belong in the UWGAC repo.

It can be argued this should get its own page as it relates to setting up a Cromwell configuration file, which could be useful for stuff besides the concurrent job limit, and so we can point users to it out of context of a tutorial... but for now this seems like a good place for this information.

### DIFF
--- a/docs/getting-started/getting-started-with-wdl.rst
+++ b/docs/getting-started/getting-started-with-wdl.rst
@@ -209,16 +209,16 @@ calling it via Docker. This will test that the WDL correctly describes
 how to run your tool.
 
 Setting Up the Dockstore CLI
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The first thing I'll do is
 `setup the Dockstore CLI locally <https://dockstore.org/quick-start>`__.
 This will have me install all of the dependencies needed to run the
 Dockstore CLI on my local machine.
 
-Although not strictly required, it's a good idea to set a concurrent-job-limit when running locally. This is because Cromwell, when running locally, loses the ability to set compute resources within what is specified in a task's runtime attributes. As such, the local version of Cromwell sometimes uses too many resources when running on scattered tasks, causing the task to get sigkilled or for Docker to lock up. Of course, there is a tradeoff: If you set concurrent-job-limit, tests involving scattered tasks may execute slower as less instances of a scattered task will run in parallel.
+Although not strictly required, it's a good idea to set a concurrent-job-limit when running locally. This is because Cromwell, when running locally, loses the ability to set compute resources within what is specified in a task's runtime attributes. As such, the local version of Cromwell sometimes uses too many resources when running scattered tasks, causing the task to get sigkilled or for Docker to lock up. Of course, there is a tradeoff: If you set concurrent-job-limit, tests involving scattered tasks may execute slower as less instances of a scattered task will run in parallel.
 
-.. tip::  If a Docker lockup happens, you will notice tasks do not progress beyond WaitingForReturnCode and you will be temporarily unable to use Docker on your OS. Thankfully, this can be solved easily by restarting Docker on your machine.
+.. tip::  If a Docker lockup happens, you will notice tasks do not progress beyond WaitingForReturnCode and you will be temporarily unable to use Docker on your OS. This can be resolved by restarting Docker on your machine.
 
 The easiest way to avoid these issues is to set up a Cromwell configuration file that sets concurrent-job-limit.
 
@@ -251,12 +251,12 @@ You may optionally edit ``root = "cromwell-executions"`` to something else if yo
 
 ::
 
-    cromwell-vm-options: -Dconfig.file=<path-to-your-cromwell-config>
+    cromwell-vm-options: -Dconfig.file=<absolute-path-to-your-cromwell-config>
 
 You're now all set up -- the Dockstore CLI will use this configuration file, and will only allow one instance of a scattered task to run at once.
 
 Set Up Local Data
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^
 
 Next thing I’ll do is create a completely local dataset and JSON
 parameterization file:
@@ -282,7 +282,7 @@ You can see in the above I give the relative path to the input under
 ``bam_input`` and the memory in GB that I want to use for the task.
 
 Run Your Tool
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^
 At this point, let's run the tool with our local inputs and outputs via
 the JSON config file:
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 The first thing I'll do is
 `setup the Dockstore CLI locally <https://dockstore.org/quick-start>`__.
 This will have me install all of the dependencies needed to run the
 Dockstore CLI on my local machine.
 
+Although not strictly required, it's a good idea to set a concurrent-job-limit when running locally. This is because Cromwell, when running locally, loses the ability to set compute resources within what is specified in a task's runtime attributes. As such, the local version of Cromwell sometimes uses too many resources when running on scattered tasks, causing the task to get sigkilled or for Docker to lock up. Of course, there is a tradeoff: If you set concurrent-job-limit, tests involving scattered tasks may execute slower as less instances of a scattered task will run in parallel.
+
+.. tip::  If a Docker lockup happens, you will notice tasks do not progress beyond WaitingForReturnCode and you will be temporarily unable to use Docker on your OS. Thankfully, this can be solved easily by restarting Docker on your machine.
```
coverbeck, line 221: Personal preference, but I would remove "Thankfully, ".

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
...
+
+The easiest way to avoid these issues is to set up a Cromwell configuration file that sets concurrent-job-limit.
+
+1. Download `this template file <https://github.com/broadinstitute/cromwell/blob/develop/cromwell.example.backends/cromwell.examples.conf>`__ and name it ``.cromwell.conf``
```
coverbeck, line 225: Is that a convention, to make it `.cromwell.conf`? If it's not a convention, I would be inclined to just call it `cromwell.conf`.

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
...
 
+Setting Up the Dockstore CLI
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 The first thing I'll do is
 `setup the Dockstore CLI locally <https://dockstore.org/quick-start>`__.
 This will have me install all of the dependencies needed to run the
 Dockstore CLI on my local machine.
 
+Although not strictly required, it's a good idea to set a concurrent-job-limit when running locally. This is because Cromwell, when running locally, loses the ability to set compute resources within what is specified in a task's runtime attributes. As such, the local version of Cromwell sometimes uses too many resources when running on scattered tasks, causing the task to get sigkilled or for Docker to lock up. Of course, there is a tradeoff: If you set concurrent-job-limit, tests involving scattered tasks may execute slower as less instances of a scattered task will run in parallel.
```
coverbeck, line 219: "running on scattered tasks" => "running scattered tasks"?

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
...
+            # The backend custom configuration.
+            config {
+                # Optional limits on the number of concurrent jobs
+                concurrent-job-limit = 1
+
+You may optionally edit ``root = "cromwell-executions"`` to something else if you wish to be certain that your configuration is getting used, as it will change the name of the executions folder. In any case, close that file and add the following line to ~/.dockstore/config
+
+::
+
+    cromwell-vm-options: -Dconfig.file=<path-to-your-cromwell-config>
```
coverbeck, line 254: Does the path need to be absolute? If yes, then I suggest `<absolute-path-to...`

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
 calling it via Docker. This will test that the WDL correctly describes
 how to run your tool.
 
+Setting Up the Dockstore CLI
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
coverbeck, line 212: Line 212's length should match line 211's to be consistent with other files in the project.

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -208,11 +208,56 @@ the Dockstore command line and descriptor rather than just directly
...
+You may optionally edit ``root = "cromwell-executions"`` to something else if you wish to be certain that your configuration is getting used, as it will change the name of the executions folder. In any case, close that file and add the following line to ~/.dockstore/config
+
+::
+
+    cromwell-vm-options: -Dconfig.file=<path-to-your-cromwell-config>
+
+You're now all set up -- the Dockstore CLI will use this configuration file, and will only allow one instance of a scattered task to run at once.
+
+Set Up Local Data
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
coverbeck, line 259: Same length at 258

docs/getting-started/getting-started-with-wdl.rst
```diff
@@ -236,6 +281,8 @@ it really doesn't matter. I'm using a sample I checked in already:
 You can see in the above I give the relative path to the input under
 ``bam_input`` and the memory in GB that I want to use for the task.
 
+Run Your Tool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
coverbeck, line 285: Same length as 284.

coverbeck (approved):

docs/getting-started/getting-started-with-wdl.rst
denis-yuen: Don't feel strongly, but could also consider making things a bit simpler by always picking a specific location like `.dockstore/cromwell.conf`

denis-yuen (approved):

avani-k (approved):

docs/getting-started/getting-started-with-wdl.rst
aofarrel: My understanding was that the convention is .cromwell.conf because that's what the Cromwell jar will look for by default. In the context of the Dockstore CLI it's a moot point as we have to point to it directly anyway, but I think keeping the convention will help people who may switch runners. (That said, if we set up the Dockstore CLI to come with the template file itself, I do think `.dockstore/` is a good place for it.)

But now that I look again, it seems that maybe the jar would need to be passed the file anyway? https://cromwell.readthedocs.io/en/stable/tutorials/ConfigurationFiles/

docs/getting-started/getting-started-with-wdl.rst
aofarrel: Good spot. I think it can be relative, but mine is absolute and there is no reason not to make that the standard.

coverbeck:
I checkout the PR locally, and ran `make html` and `make linkcheck` locally, and they both passed. Merging with admin privileges.

PR #133, opened 2021-09-22
Feature/update launch with galaxy

avani-k:


docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,8 +2,7 @@
 Getting Started with Galaxy
 ===========================
 
-Galaxy workflow support in Dockstore is new and a preview of
-more full featured support to come.
+Dockstore has recently launched the Galaxy workflow support and the _`Launch with Galaxy <../launch-with/galaxy-launch-with>`__ tutorial elaborates different ways to launch a Galaxy workflow.
```
coverbeck, line 5: I suggest you rephrase to not have "recently launched" -- otherwise somebody is going to have come back here a year from now and remove "recently" (and remember to do it!).

coverbeck:
Visible here (I find it easier to review that way, sometimes): https://docs.dockstore.org/en/feature-update-launchwithgalaxy/

docs/getting-started/getting-started-with-galaxy.rst
coverbeck: Link not showing up correctly in https://docs.dockstore.org/en/feature-update-launchwithgalaxy/getting-started/getting-started-with-galaxy.html

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,8 +2,7 @@
...
 
 Unlike WDL and CWL, Galaxy
 workflows in the near term are primarily created and modified
```
coverbeck, line 8: This comment is for line 13/12 (I can't comment on directly, because you didn't change it) -- you should remove the part where it says you can't launch with.

coverbeck (changes requested):

docs/launch-with/galaxy-launch-with.rst
```diff
@@ -4,23 +4,30 @@ Galaxy
...
 
-The button will bring up a dropdown window where you may select which Galaxy instance to launch the workflow into. A few default options will
-be provided by Dockstore, but if you need to launch into another instance, you may provide a custom URL. For your custom URL to work properly,
+The button will bring up a dropdown window where you may select which Galaxy instance to launch the workflow into. A few default options are provided by Dockstore, but if you need to launch into another instance, you may provide a custom URL. For your custom URL to work properly,
 it must start with ``https://`` and there should not be an ending forward slash ``/``.
 
 .. figure:: /assets/images/docs/galaxy/launch_with_options.png
    :alt: Launch options for Galaxy
+   
+You can use Terra to spin up a custom cloud environment for Galaxy. To do that, navigate to the Notebooks tab on any Workspace on `Terra <https://app.terra.bio/>`__ and click on the Create a Cloud Environment for Galaxy on the left.
```
denis-yuen, line 25: Appearance of Terra seems a bit abrupt since we introduce usegalaxy.org on line 5. 
Thinking this needs a header like "While browsing Galaxy", for example "While using Terra"
and should be moved below "While browsing Galaxy" to complete the thought about usegalaxy.org before moving onto Terra.

denis-yuen (changes requested):
Thanks for fleshing this out!

docs/launch-with/galaxy-launch-with.rst
avani-k: The placement of this is a bit tricky. What I was going for is that, on Dockstore, user can use the default cloud options or spin up a custom cloud environment on Terra (BDC or AnVIL included). And hence it is under the While browsing Dockstore header. The concept was to introduce 3 different ways to launch the Galaxy workflows:

1. From Dockstore, use default cloud instance or add unique Galaxy cloud instance URL (up to and including /galaxy)
2. Within Galaxy, hit import, use their search feature
3. Within Galaxy, hit import, copy TRS ID from Dockstore entry

But if the 1st way is elaborate and the introduction of Terra is abrupt, we can technically introduce that as a different way to launch Galaxy. So we can have browsing Dockstore and using the available cloud instances as one way. And introduce Terra and how to create a custom cloud instance and paste it in Dockstore as a whole other way.

I'd love to know your thoughts! @denis-yuen and @coverbeck @bethsheets

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,16 +2,12 @@
 Getting Started with Galaxy
 ===========================
 
-Galaxy workflow support in Dockstore is new and a preview of
-more full featured support to come.
+Dockstore supports Galaxy workflows and the :doc:`Launch with Galaxy <../../launch-with/galaxy-launch-with>` tutorial elaborates different ways to launch a Galaxy workflow. Please note that Galaxy workflows cannot be launched by the Dockstore CLI.
```
bethsheets, line 6: Here, I would add two sections: 

Section: Using Galaxy Servers (such as usegalaxy.org)
Institutions across the world maintain Galaxy instances that are often shared with the greater community for free, often subsidized by grants. These are great ways to learn and use Galaxy, etc. etc...

Section: Creating a Custom Galaxy in the Cloud with Terra, AnVIL, and BioData Catalyst
Galaxy is now integrated into the Terra cloud workspace and this provides Galaxy users the ability to create a custom environment that they can scale to their needs. Additionally, the Terra workspace is able to access a rich data corpus hosted in the cloud, including many controlled-access datasets. 

Something like that? Maybe expand a bit.

bethsheets (changes requested):

docs/launch-with/galaxy-launch-with.rst
coverbeck: > using Galaxy with controlled access data. It is my impression that the other available Galaxy instances are free, but users can only bring small and open datasets, and users are limited in computing resources.

Hmm, I'm not so sure about that. I infer, maybe incorrectly, that many institutions run their own Galaxy instances, and I would think those have access to controlled access data.

And we saw Galaxy is used by Australian BioCommons, which I assume will have controlled access data.

bethsheets: This is true, Charles. I think we can avoid discussing the nuance in the documentation. But to clarify, my discussion around controlled access data relates to non-members of these Galaxy servers using them with controlled access data. Generally, controlled data access requests require the researcher to work with the compute clusters to follow specific practices. It is difficult to obtain this when using a free resource that your sponsoring institute is not affiliated with.

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,16 +2,22 @@
...
+
+Galaxy instances are offered by many resources, free and otherwise. Each come with their own advantages and users can make a choice based on their own needs.
```
denis-yuen, line 7: Maybe "many institutions" rather than "many resources". 
The Galaxy instance itself seems like a resource?

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,16 +2,22 @@
...
+
+Using Galaxy Servers
+--------------------
+Institutions across the world maintain Galaxy instances that are often shared with the greater community for free, subsidized by grants, such as the `Use Galaxy Org <https://usegalaxy.org/>`. It offers a quick and customizable way to spin up a Galaxy instance. These are great ways to learn and use Galaxy, and many offer valuable learning resources such as the `Galaxy Training Network <https://training.galaxyproject.org/>`.
```
denis-yuen, line 11: "It offers a quick and customizable way to spin up a Galaxy instance." should maybe be "It offers a quick and customizable way to spin up a Galaxy workflow."

Not sure, but would have thought usegalaxy.org is a Galaxy instance itself as opposed to a way to spin up new ones.

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,16 +2,22 @@
...
+
+Custom Galaxy in the Cloud with Terra, AnVIL, and BioData Catalyst
+------------------------------------------------------------------
+Galaxy is now integrated into the Terra cloud workspace and this provides Galaxy users the ability to create a custom environment that they can scale to their needs. Additionally, the Terra workspace is able to access a rich data corpus hosted in the cloud, including many controlled-access datasets, thus allowing users to securely conduct analyses with this data.
```
denis-yuen, line 15: "Terra workspaces offer quick and customizable ways to spin up a Galaxy instance" seems to belong here.

docs/getting-started/getting-started-with-galaxy.rst
avani-k: What I mean to say is that the user can customize their Galaxy "workspace" in the sense of the compute space and resources without having to know much about the cloud backend, and hence it is quick. I am looking for another word for what I want to convey. Changing it to Galaxy workflow makes it seem like the workflow itself is customizable, which it may be, but not what I want to focus on in this section.

docs/getting-started/getting-started-with-galaxy.rst
avani-k: Ahh I see what happened here. That was a hard find!

bethsheets (approved):
Looks great!

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -2,16 +2,22 @@
...
 
-Galaxy workflow support in Dockstore is new and a preview of
-more full featured support to come.
+Dockstore supports Galaxy workflows and the :doc:`Launch with Galaxy <../../launch-with/galaxy-launch-with>` tutorial elaborates different ways to launch a Galaxy workflow. Please note that Galaxy workflows cannot be launched by the Dockstore CLI.
+
+Galaxy instances are offered by many institutions, free and otherwise. Each come with their own advantages and users can make a choice based on their own needs.
+
+Using Galaxy Servers
+--------------------
+Institutions across the world maintain Galaxy instances that are often shared with the greater community for free, subsidized by grants, such as the `Use Galaxy Org <https://usegalaxy.org/>`. These are great ways to learn and use Galaxy, and many offer valuable learning resources such as the `Galaxy Training Network <https://training.galaxyproject.org/>`.
```
coverbeck, line 11: Both the links do not show up correctly on https://docs.dockstore.org/en/feature-update-launchwithgalaxy/getting-started/getting-started-with-galaxy.html.

I think you need trailing double underscores on each link, but please verify.

coverbeck (changes requested):

coverbeck (approved):

denis-yuen (approved):

PR #141, opened 2021-11-03
Update GitHub Apps documentation

garyluu:
For dockstore/dockstore#4179

Add a TL;DR in hopes that it will be read

### DIFF
--- a/docs/getting-started/github-apps/github-apps.rst
+++ b/docs/getting-started/github-apps/github-apps.rst
@@ -6,6 +6,12 @@ Automatic Syncing with GitHub Apps and .dockstore.yml
 
 Overview
 --------
+Summary of GitHub App usage:
+
+- Install Dockstore GitHub App
+- Push a new commit with a .dockstore.yml file present
+- Wait a while for Dockstore to process the new version
+- See workflow on Dockstore
 
 This document gives a high level overview of how Dockstore uses GitHub apps.
 For extra details on configuring and using the Dockstore
--- a/docs/getting-started/github-apps/github-apps.rst
+++ b/docs/getting-started/github-apps/github-apps.rst
@@ -8,10 +8,11 @@ Overview
 --------
 Summary of GitHub App usage:
 
-- Install Dockstore GitHub App
-- Push a new commit with a .dockstore.yml file present
-- Wait up to 5 minutes for Dockstore to process the new version
-- See workflow on Dockstore
+
+#. Install Dockstore GitHub App
+#. Push a new commit with a .dockstore.yml file present
+#. Wait up to 5 minutes for Dockstore to process the new version
+#. See workflow on Dockstore
 
 This document gives a high level overview of how Dockstore uses GitHub apps.
 For extra details on configuring and using the Dockstore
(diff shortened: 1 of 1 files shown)


docs/getting-started/github-apps/github-apps.rst
```diff
@@ -6,6 +6,12 @@ Automatic Syncing with GitHub Apps and .dockstore.yml
 
 Overview
 --------
+Summary of GitHub App usage:
+
+- Install Dockstore GitHub App
+- Push a new commit with a .dockstore.yml file present
+- Wait a while for Dockstore to process the new version
```
coverbeck, line 13: Maybe "Wait up to 5 minutes"? In my experience, it takes about a minute and half, so that should give us buffer. Or 10, to be extra safe.

coverbeck (approved):

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -6,6 +6,12 @@ Automatic Syncing with GitHub Apps and .dockstore.yml
 
 Overview
 --------
+Summary of GitHub App usage:
+
+- Install Dockstore GitHub App
+- Push a new commit with a .dockstore.yml file present
```
denis-yuen, line 12: This seems too subtle and indirect. 

How about something more like "After installing the github app, make sure to push a new commit to a branch or create a tag with a .dockstore.yml file."

denis-yuen (commented):
Can we make this more obvious? 
Like with the note or important box in 
https://docs.dockstore.org/en/stable/getting-started/hosted-tools-and-workflows.html

docs/getting-started/github-apps/github-apps.rst
garyluu: the important box is for important notes, I don't think that's appropriate for summaries. "After installing the GitHub app" is already mentioned in the point above it.  "create a tag with a .dockstore.yml" seems like an edge case in which the person already has a .dockstore.yml present and doesn't need to make another commit.

garyluu: Changed to numbered list to indicate that it's a sequential process. The summary is already placed at the top of the page, if they decide to skip that, then I don't think there's anything we can do that would make a difference

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -6,6 +6,13 @@ Automatic Syncing with GitHub Apps and .dockstore.yml
 
 Overview
 --------
+Summary of GitHub App usage:
+
+
+#. Install Dockstore GitHub App
```
DailyDreaming, line 12: A  link to how to "Install Dockstore GitHub App" on the line that says that would be useful for someone like me.

DailyDreaming (approved):
LGTM.  Just one comment.

docs/getting-started/github-apps/github-apps.rst
garyluu: putting this inside the note box as mentioned with the use case in https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html . seems to make sense

denis-yuen: > the important box is for important notes, I don't think that's appropriate for summaries 

Fair, I'm ok with rewording it to not be a summary and to make it an important note.
Looking at the preview, the numbered list doesn't stand out at all

denis-yuen (changes requested):
Commented above

garyluu:
Looks like this now
image

denis-yuen (approved):

NatalieEO (approved):

PR #142, opened 2021-11-04
Update FAQ with discover button

chmreid:
Closes [SEAB-3416](https://ucsc-cgl.atlassian.net/jira/software/c/projects/SEAB/boards/72?selectedIssue=SEAB-3416&assignee=5d3f42b9f14e7c0da562927e)


docs/faq.rst
```diff
@@ -261,7 +261,10 @@ How do I add other users as maintainers of a tool?
 For tools registered on Quay.io and workflows registered with GitHub,
 Dockstore automatically allows users from the same Quay.io organization
 or GitHub organization to manage tools/workflows together (users will
-need to "Refresh Organization" or "Refresh All").
+need to click "Discover Existing Dockstore Workflows" in the "My workflows"
```
coverbeck, line 264: I know the JIRA ticket suggested this language, but it doesn't read very well to me, mainly because of the way it already was.

1. The FAQ item is titled "How do I add other users as maintainers of a tool?" But then the answer discusses both tools and workflows. I suggest either changing the question to "How do I add other users as maintainers of a tool or worklow?", or creating a new item "How do I add other users as maintainers of a workflow?".
2. If we keep them under the same item, I suggest having separate paragraphs for tools and workflows.
3. I think the case this is covering is somebody else in your GitHub org registers a workflow with GitHub apps, but it doesn't show up under your My Workflows. I feel it could be clearer.

coverbeck (changes requested):
The screenshot doesn't have the Discover Existing Dockstore Workflows popup. That needs to be addressed.

My inline comment is opening a can of worms and may go beyond the scope of this. I'm OK not acting on that and/or creating a followup ticket.

docs/faq.rst
denis-yuen: > "How do I add other users as maintainers of a tool or worklow?"

Changing the question seems good to me

> I think the case this is covering is somebody else in your GitHub org registers a workflow with GitHub apps, but it doesn't show up under your My Workflows. I feel it could be clearer. 

I believe it is

chmreid: updated question in 708e8fa

chmreid:
updated screenshot to include "Discover Workflows" button in 708e8fa

docs/faq.rst
```diff
@@ -255,13 +255,16 @@ How can I use the Dockstore CLI with Python 2?
...
-How do I add other users as maintainers of a tool?
---------------------------------------------------
+How do I add other users as maintainers of a tool or worklow?
+-------------------------------------------------------------
 
 For tools registered on Quay.io and workflows registered with GitHub,
 Dockstore automatically allows users from the same Quay.io organization
-or GitHub organization to manage tools/workflows together (users will
-need to "Refresh Organization" or "Refresh All").
+or GitHub organization to manage tools and workflows together (users will
```
coverbeck, line 263: Thanks for updating the screenshot.

I find this paragraph hard to understand. It's one long sentence, talks about tools, workflows and GitHub, but shows a screenshot that is specific to GitHub-based workflows. It doesn't explain when you would use Discover.

It may just be me; I added Natalie as a reviewer for her opinion.

coverbeck (approved):
I have my reservations about this, but approving, and I'll let others opine.

garyluu (approved):

docs/faq.rst
denis-yuen: Re-reading, I guess my comment was ambiguous. I feel that @coverbeck identified the case correctly that this is covering. I wasn't commenting on whether this could be clearer or not. (It probably could)

denis-yuen (approved):

docs/faq.rst
NatalieEO: I think this answer is pretty confusing. I'm not sure why it mentions tools, but doesn't give an answer for tools. I don't think there's even an equivalent button to "Discover Existing Dockstore Workflows" for tools?

docs/faq.rst
```diff
@@ -255,13 +255,16 @@ How can I use the Dockstore CLI with Python 2?
 Python 2 support ends in 2020. You can get the Python 2 requirements.txt with ``curl -o requirements.txt "https://dockstore.org/api/metadata/runner_dependencies?python_version=2"``
 but it is currently untested.
 
-How do I add other users as maintainers of a tool?
---------------------------------------------------
+How do I add other users as maintainers of a tool or worklow?
+-------------------------------------------------------------
 
 For tools registered on Quay.io and workflows registered with GitHub,
 Dockstore automatically allows users from the same Quay.io organization
```
NatalieEO, line 262: It was already like this, but we use the term "namespace" in our UI for tools, although Quay also refers to it as organization. Regardless, I think the answer tools and workflows needs to separated. This only answers it for workflows. I guess the process is to "refresh namespace" ?

NatalieEO (changes requested):

chmreid:
Circling back to this pull request. I've updated the question/answer to remove references to tools and only mention workflows. (Expectation is that another FAQ question covering tools can be added separately later.)

I've been struggling to understand the existing answer to this question, though. This paragraph, for example, doesn't say anything clear about workflows:

> For tools registered on Docker Hub, GitLab or private registries, this
> feature currently does not exist because these registries do not allow
> the retrieval of organization information. **Likewise, workflows
> registered with other source code repositories lack this feature.**

(I don't know what "other source code repositories" is referring to. Is it talking about the public/private repository distinction? Is it talking about GitLab/BitBucket vs GitHub, or maybe GitHub/GitLab/BitBucket vs. repository hosting sites not supported by Dockstore?)

Also, is this paragraph still relevant?

> For participants of the :doc:`limited sharing
> beta, you can enter the email
> addresses of the users you wish to share with to give them permissions
> to your workflow. This is only available for hosted workflows and users
> with Google accounts linked to Terra.

I've done my best to update, please have another look and let me know if this is an acceptable improvement to the FAQ question.

docs/faq.rst
```diff
@@ -252,23 +252,19 @@ not limited to):
...
-How do I add other users as maintainers of a tool?
---------------------------------------------------
+How do I add other users as maintainers of a workflow?
+------------------------------------------------------
 
-For tools registered on Quay.io and workflows registered with GitHub,
-Dockstore automatically allows users from the same Quay.io organization
-or GitHub organization to manage tools/workflows together (users will
-need to "Refresh Organization" or "Refresh All").
+For workflows registered with GitHub, Dockstore automatically allows users from
```
coverbeck, line 261: I suggest something like this, needs tweaking as I ran out of steam:


For workflows registered with GitHub, Dockstore allows users from the same GitHub organization to manage workflows together. If a new GitHub workflow is added to Dockstore by another user, you may need to tell Dockstore to add it to your workflows.

If the workflow was added to Dockstore via the GitHub App, click Discover Existing Dockstore Workflows.

If the workflow was added by manually registering it, click Refresh Organization.


@aofarrel Thoughts?

@denis-yuen @garyluu , in writing this suggestion, it seems like the web service could automatically do this -- when a workflow is added to Dockstore, the web service could find all the other GitHub users for that org and auto add them to the workflow. Is there a reason we don't do that already? If not, I'll create a new issue.

garyluu: We don't auto-add them to the organization because then the other users complain that their my-workflows and my-tools is bloated (and we're missing a search to filter those out)

NatalieEO: > If the workflow was added to Dockstore via the GitHub App, click Discover Existing Dockstore Workflows.
If the workflow was added by manually registering it, click Refresh Organization.

I don't think this is right...I think the Discover existing workflow is used when you don't see an entire organization listed and doesn't really have anything to do with the type of workflow it is.

NatalieEO: https://github.com/dockstore/dockstore-ui2/pull/984

coverbeck: > 

Aargh, good catch, you're right. I'd seen [this code](https://github.com/dockstore/dockstore/blob/477977c5e657c3c2339f64da6e6f905ba5513aa5/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L943) and somehow convinced myself the Discover button was calling that.

coverbeck: So my suggestion becomes something like:

For workflows registered with GitHub, Dockstore allows users from the same GitHub organization to manage workflows together. If a new GitHub workflow from the same GitHub organization is added to Dockstore by another user, click the Discover Existing Dockstore Workflows button so it will appear in My Workflows.

denis-yuen: @denis-yuen @garyluu  

I think we can consider auto-add and discuss  in a new ticket. Not sure avoiding this is the best way of avoiding bloat (wouldn't they get it eventually if they ever hit discover?)  A filter or "I am not interested" marker is reasonable too.  

That said, it doesn't avoid the need for a discover button or this FAQ since other events can occur that change access without a workflow being added to Dockstore. (Being added to an org, being given access to a repo, etc.)

denis-yuen (approved):
With suggested improvements in wording

NatalieEO (changes requested):
I also would prefer Charles' suggestion

chmreid:
@NatalieEO @coverbeck Updated wording and re-requested review, please let me know if this final version is consistent with what Charles suggested.

NatalieEO (approved):

PR #151, opened 2021-12-07
Overhaul GitHub app faqs

aofarrel:
Remove out of date information, reorganize, and clarify. See commit comments for more information.


docs/getting-started/github-apps/github-apps.rst
```diff
@@ -270,28 +293,7 @@ filters
...
-compare it with our examples/documentation to confirm that you filled it in correctly. In the future we plan to have a system in place where
-users can keep track of these GitHub events and resulting action taken by Dockstore, even if the message was succesfully handled.
-
-Another possible issue is that we received the message from GitHub, but the user who triggered the message event is not registered on Dockstore with
-the corresponding GitHub account. This is only an issue if the workflow or service does not already exist on Dockstore. When creating new workflows and
-services, we need to be able to associate them with a user. If the workflow or service already exists on Dockstore, then this error will not occur and the 
-version will be properly added/updated/deleted on Dockstore.
-
-As always, you can reach out to our team on our `discussion forum <https://discuss.dockstore.org/>`_ to discuss any issues you are facing.
+For more info on services and registering them, check out our :doc:`Getting Started with Services </getting-started/getting-started-with-services>`, or our :doc:`Service Version 1.2 Template</assets/templates/template.html#service-version-1-2-template>`.
```
coverbeck, line 296: Do you need the part starting with `or our :doc:...`? If so, it's not a valid link and gives a build warning.

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -66,6 +66,29 @@ to associate with a workflow on Dockstore. A template for both workflows and ser
 along with explanations for each field. For every branch/release on GitHub that has one of these files, a corresponding entry
 will be made on Dockstore.
 
+Error Handling
+----------------------------------
+Since Dockstore relies on GitHub to tell us when changes have been made on GitHub, there are chances that the message gets lost or delayed.
+Typically, Dockstore reacts within seconds of a change being made on GitHub, however service disruptions can delay this to a few minutes.
```
coverbeck, line 72: "within seconds" -> "within a couple of minutes". Last I checked, it was about 90 seconds, 2 minutes leaves us some buffer room.

GitHub does notify us right away I believe, but we have an intentional delay in there due to some issues we ran into. We should not set expectations that the users will see changes in Dockstore almost immediately.

coverbeck: I would just drop the service disruptions section.

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -66,6 +66,29 @@ to associate with a workflow on Dockstore. A template for both workflows and ser
...
+If a message were to get lost, unfortunately you will need to push to GitHub again. Currently, there is no way to tell on Dockstore whether
```
coverbeck, line 73: I'd drop this paragraph.

If we don't get  the message from GitHub, which I suspect is very rare, we wouldn't know that we didn't get it. :) So nothing is going to change in the future.

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -66,6 +66,29 @@ to associate with a workflow on Dockstore. A template for both workflows and ser
...
+a GitHub message was delayed or lost. We recommend waiting a few minutes and then trying to push again. This will be changed in the future.
+
+Another error that could occur is that we received the message from GitHub, however the ``/.dockstore.yml`` is invalid. If we cannot read the
```
coverbeck, line 76: If this happens, you may see an error in GitHub app logs. I think it depends how bad it is. :) If the file doesn't exist at all (or has a typo in its name), then you won't get anything in the logs. But I believe if there is a .dockstore.yml file, you'll always see something in GitHub app logs.

docs/getting-started/dockstore-workflows.rst
```diff
@@ -114,31 +114,25 @@ This is the newest way of getting content onto Dockstore and is by far the most
...
+   :width: 40 %
 
    Install our GitHub App on either all repositories in an organization or on specific repositories
 
 Once you've installed our GitHub app on a repository or organization, you'll need to add a ``/.dockstore.yml`` file to
 the root directory of a branch of the repository that contains your workflow. This file contains information like
 workflow path, test parameter file, workflow name, etc. When a push is made on GitHub to a branch
 with a ``/.dockstore.yml``, Dockstore will add that branch to the corresponding workflow on Dockstore. If the
-workflow doesn't already exist on Dockstore, one will be created.
+workflow doesn't already exist on Dockstore, one will be created. Note that if you have more than workflow in your repo, you can describe multiple in one ``/.dockstore.yml`` file to create multiple workflow entries.
```
NatalieEO, line 135: typo/suggestion: more than **one** workflow, you can describe each one in a single ``/.dockstore.yml``

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
```diff
@@ -16,46 +16,55 @@ This means less manual work for workflow developers, and less waiting for conten
 
 This requires the addition of a ``/.dockstore.yml`` file to your repository on GitHub.
 This file contains workflow information such as workflow path, test parameter paths, etc. that Dockstore will use to setup
-the corresponding workflow on Dockstore. It's important to note, that you will need a ``/.dockstore.yml`` file on each branch of your GitHub
+the corresponding workflow on Dockstore. Branches that do not have a ``/.dockstore.yml`` file on each branch of your GitHub
```
NatalieEO, line 19: ?

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -66,6 +66,29 @@ to associate with a workflow on Dockstore. A template for both workflows and ser
 along with explanations for each field. For every branch/release on GitHub that has one of these files, a corresponding entry
 will be made on Dockstore.
 
+Error Handling
```
NatalieEO, line 69: I know you only moved this section, but I think this should really mention to look at the GitHub app logs as one of the first things to do.

NatalieEO (changes requested):

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
```diff
@@ -16,46 +16,55 @@ This means less manual work for workflow developers, and less waiting for conten
...
 -------------------------------------------
-Adding a ``/.dockstore.yml`` file to a template branch (ex: master, develop, main), will make it so
+Adding a ``/.dockstore.yml`` file to a template branch (ex: master, develop, main) will make it so
 any new branches created from this template will be automatically added to and synced on Dockstore.
 
 Therefore, as long as your workflow is already registered on Dockstore and your ``/.dockstore.yml`` is configured correctly,
 then updates to the workflow (including adding new versions) should happen continuously and automatically.
 
-For this setup, if you *do not* want a new GitHub branch to generate a corresponding workflow-version on Dockstore,
+On the other hand, if you *do not* want a new GitHub branch to generate a corresponding workflow-version on Dockstore,
```
coverbeck, line 32: This sentence was written before we added filters to the .dockstore.yml, and that would arguably be the preferred way of handling this -- it's going to be a hassle to merge back to the main branch if you deleted the .dockstore.yml in the feature branch.

If you feel like tackling this here, that would be great. If not, we can create another ticket for followup.

docs/getting-started/dockstore-workflows.rst
```diff
@@ -114,31 +114,25 @@ This is the newest way of getting content onto Dockstore and is by far the most
...
-.. figure:: /assets/images/docs/gh-app-reg-1.png
-   :alt: GitHub App Registration on Dockstore
+.. image:: /assets/images/docs/add-workflow-button.png
+   :width: 40 %
 
-   Select Register Using GitHub Apps to get a link to our installation page
+.. image:: /assets/images/docs/register-workflow-github-apps.png
+   :width: 40 %
 
-.. figure:: /assets/images/docs/gh-app-reg-2.png
```
coverbeck, line 127: If you're not going to display this screenshot (is that intentional?), then please remove the .png as well.

docs/getting-started/dockstore-workflows.rst
```diff
@@ -114,31 +114,25 @@ This is the newest way of getting content onto Dockstore and is by far the most
 GitHub Apps, Dockstore can react to changes on GitHub as they are made, keeping Dockstore synced
 with GitHub automatically.
 
-To install our GitHub App on either a repository or organization,
-navigate to the ``/my-workflows`` page and click add workflow. Follow the steps for GitHub Apps and
-you will be redirected to GitHub where you can select which repositories to install the Dockstore
-GitHub app on. The process is shown in the following images.
+Installing the GitHub App is simply -- navigate to ``/my-workflows``, click the ``+`` button on the left hand sidebar, select ``Register using GitHub Apps``, and then click
+``+ Manage Dockstore Installation on GitHub``. You'll then be redirected to GitHub where you can select which repositories can be accessed by the GitHub app.
```
coverbeck, line 118: Nit-picky, and maybe this is fine the way it is, but you are granting the GitHub app permissions to access your repo(s). I feel "select which... can be accessed" is a little passive.

docs/getting-started/dockstore-workflows.rst
coverbeck, line 135: Suggest: "to create multiple workflow entries" -> "to create multiple workflows".

docs/getting-started/dockstore-workflows.rst
```diff
@@ -165,10 +159,6 @@ Now that your workflow has been added, any time there is a push to a branch on G
 it is automatically updated on Dockstore! Anytime there is a deletion of a branch on GitHub that has a ``/.dockstore.yml``, the version is
 removed from Dockstore.
 
-.. tip:: Since the workflows field is an array, this file supports multiple workflows on Dockstore stemming from
```
coverbeck, line 168: Why did you remove this? Not saying it was wrong to, but just wondering.

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
```diff
@@ -16,46 +16,55 @@ This means less manual work for workflow developers, and less waiting for conten
...
-.. image:: /assets/images/docs/github-repo-settings.png
+.. image:: /assets/images/docs/reinstall-app-to-cheese-org-settings.png
+   :width: 50%
+
+The changes made to my GitHub repo aren't appearing on Dockstore, but I've already installed the GitHub app and made the ``.dockstore.yml`` file. How can I figure out what's going wrong?
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+If you don't see changes, try waiting a couple of minutes and refreshing the browser on the My Workflows page again. 
 
-- Double check the ``/.dockstore.yml`` file.
+Be aware that a service disruption can result in a connection between GitHub and Dockstore getting delayed or lost. If the signal was lost, another push must be made.
```
coverbeck, line 56: This is extremely rare; I would make this the last item.

docs/getting-started/dockstore-workflows.rst
aofarrel: Because of the other elements right under it, this renders as multiple colored boxes stacked on top of each other. It's a bit distracting, and since this gives information that isn't very important, it shouldn't get its own box in a place where there's already boxes. I moved its information to line 135.

docs/getting-started/github-apps/github-apps.rst
aofarrel: This is older text that I hard time placing -- hearing that it's actually not really needed is great and will help shorten this section to the essentials. I'll bulldoze it. 🚧

docs/getting-started/github-apps/github-apps.rst
aofarrel: I'm trying to link to https://docs.dockstore.org/en/develop/assets/templates/template.html#service-version-1-2-template, what is the correct way to do so?

Additionally, we are in an awkward place where we have about four template dockstore.yml files of varying levels of complexity scattered across the docs. I was thinking of changing this template page I'm trying to link to into subpages, one for workflow .dockstore.yml, and one for service .dockstore.yml. It would not only give internal consistency but would also give a quick page for people to go to if they just need a .dockstore.yml template instead of all the info around it. What are your thoughts?

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
aofarrel: Forgot to save the edit to the second half the sentence. Intent was:
`Branches that do not have a ``/.dockstore.yml`` file will not be synchronized.`

Reason for changing was that previous wording implied that if you wanted _any_ add'l branches to be sync'd, you might need _all_ of them to have a yml, ie, if you wanted /develop/ and /master/ on GATK you would need to write hundreds of yml files including stale branches you don't care about.

docs/getting-started/dockstore-workflows.rst
aofarrel: Good spot, and a good opportunity to make it even more explicit. How about this?
`Note that a single ``/.dockstore.yml`` file can describe multiple workflows, if all of those workflows are in the same repository. This will create multiple workflow entries on Dockstore from one ``/.dockstore.yml`` file, one entry for each workflow.`

docs/getting-started/github-apps/github-apps.rst
coverbeck: Note that our live doc is on stable, not develop (just in case you're passing around https://docs.dockstore.org/en/develop links :) )

I think you want to use `:ref:` -- take a look at other examples.

docs/getting-started/dockstore-workflows.rst
aofarrel: Yes, that's intentional, I replaced it with the newer screenshots from the migration page.

aofarrel:
Open questions to reviewers:
1. Is [this](https://github.com/dockstore/dockstore-documentation/pull/151/commits/3a62af42fb13629ccf5ade4f32554f0846494912) sufficient to account for the complexities of GitHub App permissions? I do believe in the long-term we should clarify intent (see [#4613 on Dockstore](https://github.com/dockstore/dockstore/issues/4613)), but I think what I have might be sufficient to fix the comment issues users run into. And they are indeed running into it -- our best practices tell users to use GitHub orgs, so we need to prepare for the permission complexities that entails.
2. Should the .dockstore.yml workflow template be moved to the template section, alongside its services counterpart? Should it get its own page?
3. If yes to the above, should it be rewritten in the same style as the services template, ie information conveyed via comments?

docs/assets/templates/.dockstore.yml
```diff
@@ -8,13 +8,15 @@ service:
   name:
 
   # Optional array of authorship information, requires at least the name of each author.
+  # Note that just using "author" as a singular string is deprecated as of 1.2
```
coverbeck, line 11: I wouldn't bother mentioning this; I just checked and we don't have any version 1.1 .dockstore.yml

docs/getting-started/dockstore-workflows.rst
```diff
@@ -114,31 +114,25 @@ This is the newest way of getting content onto Dockstore and is by far the most
 GitHub Apps, Dockstore can react to changes on GitHub as they are made, keeping Dockstore synced
 with GitHub automatically.
 
-To install our GitHub App on either a repository or organization,
-navigate to the ``/my-workflows`` page and click add workflow. Follow the steps for GitHub Apps and
-you will be redirected to GitHub where you can select which repositories to install the Dockstore
-GitHub app on. The process is shown in the following images.
+Installing the GitHub App is simply -- navigate to ``/my-workflows``, click the ``+`` button on the left hand sidebar, select ``Register using GitHub Apps``, and then click
+``+ Manage Dockstore Installation on GitHub``. You'll then be redirected to GitHub where you can grant the app access to specific repositories within whatever organization you are installing into.
```
coverbeck, line 118: Nit-picky and you can leave it, but I think, for example, github.com/aofarrel is a user, not an org. And you make that distinction elsewhere.

docs/getting-started/dockstore-workflows.rst
```diff
@@ -114,31 +114,25 @@ This is the newest way of getting content onto Dockstore and is by far the most
...
+   :width: 40 %
 
    Install our GitHub App on either all repositories in an organization or on specific repositories
 
 Once you've installed our GitHub app on a repository or organization, you'll need to add a ``/.dockstore.yml`` file to
 the root directory of a branch of the repository that contains your workflow. This file contains information like
 workflow path, test parameter file, workflow name, etc. When a push is made on GitHub to a branch
 with a ``/.dockstore.yml``, Dockstore will add that branch to the corresponding workflow on Dockstore. If the
-workflow doesn't already exist on Dockstore, one will be created.
+workflow doesn't already exist on Dockstore, one will be created. Note that a single ``/.dockstore.yml`` file can describe multiple workflows, if all of those workflows are in the same repository. This will create multiple workflow entries on Dockstore from one ``/.dockstore.yml`` file, one entry for each workflow.
```
coverbeck, line 135: I don't like "entry" here, especially "one entry for each workflow". A workflow is an entry; this makes it sound like there's an entry and a workflow. I may be missing something though.

docs/getting-started/github-apps/github-apps.rst
```diff
@@ -174,124 +192,7 @@ entries of the `workflows` array. For each unique name present, an entry will be
...
-users can keep track of these GitHub events and resulting action taken by Dockstore, even if the message was succesfully handled.
-
-Another possible issue is that we received the message from GitHub, but the user who triggered the message event is not registered on Dockstore with
-the corresponding GitHub account. This is only an issue if the workflow or service does not already exist on Dockstore. When creating new workflows and
-services, we need to be able to associate them with a user. If the workflow or service already exists on Dockstore, then this error will not occur and the 
-version will be properly added/updated/deleted on Dockstore.
-
-As always, you can reach out to our team on our `discussion forum <https://discuss.dockstore.org/>`_ to discuss any issues you are facing.
+A template .dockstore.yml file for registering services, with explanations in the comments, can be found in our `Service Version 1.2 Template <https://docs.dockstore.org/en/develop/assets/templates/template.html#service-version-1-2-template>`_. For more info on services and registering them, check out our :doc:`Getting Started with Services </getting-started/getting-started-with-services>`.
```
coverbeck, line 195: We should not have hard-coded absolute links to the docs site; links should be relative.

coverbeck (changes requested):

NatalieEO (approved):

docs/assets/templates/.dockstore.yml
aofarrel: It is in assets/templates. Is that intentional? I don't think it gets rendered anywhere on our site, but it's still kicking around.

docs/getting-started/dockstore-workflows.rst
aofarrel: I'm using "entry" to mean a single entry on Dockstore, which consists of a workflow/tool (not a tool for .dockstore.yml registration of course), its sample JSON if any, etc. If you have 10 workflows in a repo, but only reference one in your .dockstore.yml, the other nine workflows will not become entries.

coverbeck: A workflow or tool in Dockstore already consists of its descriptors, JSON, etc.

We use the term "entry" to mean either a Dockstore Tool or a Dockstore workflow. If you go to the logged-in home page, the sidebar on the left shows "My Entries", and it will show a mixture of tools and workflows.

"One entry for each workflow" doesn't sound right to me.

aofarrel: I believe we are talking about the same thing. The only distinction I'm making is when I say 
> If you have 10 workflows in a repo, but only reference one in your .dockstore.yml, the other nine workflows will not become entries.

I mean workflow as in a WDL workflow, or a CWL workflow, etc -- which may not necessarily be on Dockstore.

docs/getting-started/dockstore-workflows.rst
aofarrel: I struggled with wording it concisely. "You'll then be redirected to GitHub where you can grant the app access to specific repositories within whatever organization or user you are installing into." is one option, but I'm wondering if that might confuse people as it may imply you can be a member of a user. Also, iirc, we are already encouraging people to have their workflows on GitHub organizations rather than personal repos (I may be confusing this with advice regarding Dockstore organizations and collections though).

docs/assets/templates/.dockstore.yml
coverbeck: Ah, I see. Yeah, I don't see any references to .dockstore-1-1.yml, so I think you can still drop this.

docs/getting-started/dockstore-workflows.rst
coverbeck: I think you can leave it as you have it then. I did say "nit-picky" :)

docs/getting-started/dockstore-workflows.rst
coverbeck: I still don't like "entry for workflow". I know "workflow" is an overloaded word, but this is the Dockstore documentation, and a workflow in Dockstore means "descriptor files, JSON", etc.. How about something like:

```
Note that a single /.dockstore.yml file can describe multiple workflows in a single repository.
```

docs/getting-started/dockstore-workflows.rst
aofarrel: I strive to remove all nits, but this one seems resistant to the metaphorical pick.

aofarrel:
> 



> When I build, there is an error/warning below. Are you expecting the text " Install our GitHub App on either all repositories in an organization or on specific repositories" to appear, I don't see it.
> 
> ```
> /Users/charlesoverbeck/git/dockstore/dockstore-documentation/docs/getting-started/dockstore-workflows.rst:126: WARNING: Error in "image" directive:
> no content permitted. I don't see the text 
> 
> .. image:: /assets/images/docs/gh-app-reg-1.png
>    :width: 40 %
> 
>    Install our GitHub App on either all repositories in an organization or on specific repositories
> ```

I did some digging and it turns out that RST makes a distinction between a figure and an image. I'm not sure why, since it seems a figure is just an image with a caption (why not make the caption syntax available to images?) but I'm sure there's a reason for it. develop has these as figures, and they render with captions.
Screenshot 2021-12-15 at 9 58 37 AM
Screenshot 2021-12-15 at 10 00 39 AM

Adding or removing captions from images will cause them to not render. (This is not a meaningful whitespace issue either; both develop and my branch go /n/n/s/s/scaption, not /n/s/s/s/n/s/s/scaption.)
Screenshot 2021-12-15 at 10 11 27 AM
Screenshot 2021-12-15 at 10 11 36 AM


In the next commit, I will change the last one back into a figure to get the caption back. I also will make it a bit bigger to better "balance" this collection of three images.

coverbeck (approved):

PR #153, opened 2021-12-15
Turn warnings into errors during make

aofarrel:
***If this is merged before docker_instructions.md is turned into a format compatible with rst, it will break develop.***

`-W` turns warnings into errors
`--keep-going` will continue building even after the first error/warning

This affects `make html` both locally and [on CircleCI](https://app.circleci.com/pipelines/github/dockstore/dockstore-documentation/745/workflows/436a7d17-6948-4cd4-83a2-14900e1da4e1/jobs/761).

#154 fixes docker_instructions.md and should be merged before this.

PR #154, opened 2021-12-15
Fix markdown file's headers

aofarrel:
This fixes the warnings we've been getting during `make html`. This works with both our existing version of recommonmark and with #155. In other words this is not dependent on, but it is compatible with, #155.

This should be merged *before* #153.


garyluu (approved):

aofarrel:
It turns out this wasn't a warning after all -- it's an error. We have apparently should have been erroring out during `make` for quite some time. https://github.com/sphinx-doc/sphinx/issues/9979

NatalieEO (approved):

PR #158, opened 2022-01-20
Update ORCID.rst

avani-k:
1. Added a sentence to ensure user verifies the email associated with ORCID
2. Corrected an unfortunate typo

### DIFF
--- a/docs/end-user-topics/ORCID.rst
+++ b/docs/end-user-topics/ORCID.rst
@@ -18,7 +18,9 @@ The process is quite simple. If you are logged in now, you can `click here`_ to 
 From here, you'll be redirected to ORCID. With just a few clicks, you will now see this on the link services page of Dockstore.
 
 .. image::  /assets/images/docs/orcid_after_account_linkage.png
-  :alt: After linking ORCID to Dockstore there will be options to relink, unlick, or view your token
+  :alt: After linking ORCID to Dockstore there will be options to relink, unlink, or view your token
+
+Note: Make sure to verify the email you have registered with ORCID for this process to work.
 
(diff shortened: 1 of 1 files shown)


docs/end-user-topics/ORCID.rst
```diff
@@ -18,7 +18,9 @@ The process is quite simple. If you are logged in now, you can `click here`_ to
 From here, you'll be redirected to ORCID. With just a few clicks, you will now see this on the link services page of Dockstore.
 
 .. image::  /assets/images/docs/orcid_after_account_linkage.png
-  :alt: After linking ORCID to Dockstore there will be options to relink, unlick, or view your token
+  :alt: After linking ORCID to Dockstore there will be options to relink, unlink, or view your token
+
+Note: Make sure to verify the email you have registered with ORCID for this process to work.
```
coverbeck, line 23: I thought this part of the process did work when we looked at it? That it only failed when you tried to export to ORCID? 

Even so, I would suggest putting this note as a paragraph before line 11, e.g., something like:

Before linking your ORCID account to Dockstore, make sure you have verified the email you registered with ORCID.

coverbeck (approved):

docs/end-user-topics/ORCID.rst
avani-k: I debated about the position of this note. I will move it before line 11. The process did work, but only after verifying the email. I thought it might be useful for users to know that.

coverbeck: Sorry, I wasn't clear. I know it failed eventually and was fixed by verifying your email. I meant that your note is next to linking an ORCID account and linking an ORCID account seems to work regardless of whether you have verified your email or not. So when you said "*this* process", it sounded like you were referring to the linking, which does work.

Agreed that it's useful for users to know.

denis-yuen (approved):

PR #160, opened 2022-01-25
Reorganize FAQ page

aofarrel:
I was trying to figure out what information does and does not belong on our FAQ page and realized I was struggling to keep track of the FAQ page as it currently exists. So, I made a version that divides it into sections. There are other changes I'm playing around with, but I want to propose this one first. I made sure to keep questions that have an `.._internal-identifier`, such as the one on verified workflows, together with that identifier, and added identifiers for the sections to create a bogus TOC (bogus in that it won't get double-indexed on the sidebar).

Would like some feedback for categories. I don't like the current miscellaneous one that I have, feels kind of like a cop out. Would also like feedback r/e having all questions on the top of the page instead of just the categories, which would probably be most efficient to do via the actual TOC command (there is a workaround to keep it from overtaking the sidebar, I believe, but apparently it's a bit finicky).

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,6 +1,201 @@
 FAQ
 ===
 
+General Dockstore Questions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+What environment do you test tools in?
+--------------------------------------
+
+Typically, we test running tools in Ubuntu Linux 16.04 LTS on VMs in
+`OpenStack <https://www.openstack.org/>`__ with 8 vCPUs and 96 GB of RAM
+and above. We've also begun testing on Ubuntu 18.04 LTS and so far it's
+been successful. If you are only listing and editing tools, we have
+achieved success with much lower system requirements. However, launching
+tools will have higher system requirements dependent on the specific
+tool. Consult a tool's README or CWL/WDL description when in doubt.
+
+
+.. _what-is-a-verified-tool-or-workflow:
+
+What is a verified tool or workflow?
+------------------------------------
+
+A verified tool/workflow means that at least one version has been verified to be successfully ran on a platform.
+
+See :doc:`/advanced-topics/verification` for full details on this feature.
+
+
+What is a default version of a tool or workflow?
+------------------------------------------------
+
+Every tool/workflow is recommended to have a default version set by its
+author. It indicates to the end users which version of the tool/workflow
+they should use. For tools, the default version is uniquely identified
+by the tag from the Docker image repository. For workflows, the default
+version is identified by the Git Reference (which could be a Git tag or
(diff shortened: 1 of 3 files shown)


bethsheets (approved):
This all seems fine to me, but please also request a review by Seabright/OICR.

docs/faq.rst
```diff
@@ -1,23 +1,109 @@
 FAQ
 ===
 
-What happens if I rename my GitHub repository?
-----------------------------------------------
+Contents:
```
coverbeck, line 4: I don't see this style of TOC at the in any of other pages, although maybe I'm wrong. And it looks a lot different in a local build.

I feel you should either get rid of this section, or we should add it to all our pages, probably the former. For better or for worse, I think we should try to conform to the ReadTheDocs style.

coverbeck (commented):
I'm skipping a review of the text, since I think all you did was move it around, not edit it, and I don't think you're looking for feedback in that area. But let me know.

I do like the categories you added (but only in the sidebar :) ). It would be good to not have "Other", but coming up with names might be tricky. How about:

* Authoring and Running (creating CWL, creating Dockerfile,  more space)
* Publishing (too many versions, adding version to collection)

The any last tips one is hard. :)

docs/faq.rst
aofarrel: Contents doesn't really make sense for most pages, since you know what will be in them from the title. But an FAQ page by definition will have a hodgepodge of questions about a variety of topics (unless there are multiple FAQ pages, each focused on a specific topic, which is also an option). People don't want to have to scroll through the entire page to find out if their question may or may not be answered.

aofarrel:
@coverbeck How do you feel about this version? It is sticking to ReadTheDocs style (uses a built-in) while still being helpful to users who don't want to scroll the whole thing looking for a specific question. I managed to make it so that the headings do not appear in the sidebar unless you click the FAQ page.
Screen Shot 2022-02-18 at 3 16 35 PM

coverbeck:
The FAQ looks good to me, despite my previous objections.

avani-k (commented):
This looks great Ash! I have a suggestion for the headers. What do you think of these, I am listing just general topics not actual header names. You already have:
Dockstore CLI, GitHub, General Dockstore and so on.
How about adding either Tools or Technical Questions or something on those lines.

This PR made me think, we should probably have a longer FAQ covering a broader range of questions. I am not sure yet how we would hunt for the right topics to add to FAQs. Currently, it seems a bit heavy on Tools related questions. If we decide to add more questions, we could make the headers more specific too. Such as WDL related, CWL related and so on.

avani-k (approved):

coverbeck (approved):

PR #163, opened 2022-01-27
Give new users a simple overview of what tools, wfs, and services are

aofarrel:
A while back, I rewrote one of our docs explaining the difference between a tool and a workflow. That doc was comprehensive and therefore complex, but it's in the getting started section, which is aimed at new users. Even worse, it was at the very top of the getting started section, throwing users into the deep end immediately.

This PR:
* [Introduces a brand new* simple description of what a tool is, what a workflow is, and what a service is](https://github.com/dockstore/dockstore-documentation/blob/4455e18cb3bd6a9539cd9649ecc7035d7b6a6ae6/docs/getting-started/intro-to-dockstore-tools-and-workflows.rst)
* Moves [the detailed tool-vs-workflow stuff](https://github.com/dockstore/dockstore-documentation/blob/4455e18cb3bd6a9539cd9649ecc7035d7b6a6ae6/docs/advanced-topics/tools-vs-workflows.rst) into advanced topics
* Changes some titles to be clearer when they show up on the sidebar and in search
* Adjusts the sidebar and index.html TOC accordingly

*(except the services description, that's copied from our existing page on services)

I'm seeking input on the services description especially -- it's just one sentence for now, which is concise, but not balanced with the description of the tool and the workflow.

Note: I've mentioned wanting to make some changes to the WDL tutorial to a few people; that'll be in another PR.

PR #166, opened 2022-02-01
Modify getting started tutorial -- now with the correct commits!

aofarrel:
This is the same as #163 but it doesn't have the extra commits from develop added by mistake. I had previously based that branch on develop, then changed the branch I was merging into, so it was picking up a lot of stray commits.

### DIFF
--- a/docs/advanced-topics/tools-vs-workflows.rst
+++ b/docs/advanced-topics/tools-vs-workflows.rst
@@ -41,8 +41,6 @@ Dockstore does not support registration of these tools.
 +------------------------+------------------------------------------+-------------------------------------------------+
 | Dockstore Support      | Tool                                     | Workflow                                        |
 +========================+==========================================+=================================================+
-| Desciptor File         |  Just one file (ie, one .wdl file)       |  Can be multiple files (ie, ten .wdl files)     |
-+------------------------+------------------------------------------+-------------------------------------------------+
 | "Launch with" support  |  Only via Dockstore CLI                  |  Dockstore CLI & "launch with" buttons on entry |
 +------------------------+------------------------------------------+-------------------------------------------------+
 | Versioning             |  Based off of image's tags               |  Based off of branches/tags from Git repository,|
--- a/docs/advanced-topics/tools-vs-workflows.rst
+++ b/docs/advanced-topics/tools-vs-workflows.rst
@@ -1,9 +1,9 @@
-Tools versus Workflows: Detailed Comparison
-===========================================
+Tools vs Workflows: Detailed Comparison
+=======================================
 
 
-Comparison of Tools and Workflows
----------------------------------
+Overview
+--------
 
 When Dockstore was created, CWL was the first descriptor language we supported. CWL has a very clear distinction between a tool and a workflow.
 However, our definition for each does not completely align with the language's specification.
--- a/docs/advanced-topics/tools-vs-workflows.rst
+++ b/docs/advanced-topics/tools-vs-workflows.rst
@@ -1,50 +1,71 @@
 Tools vs Workflows: Detailed Comparison
 =======================================
 
+.. tip::
+  Generally speaking, if you care more about the Docker image than the description file, and are using CWL, you are dealing with a tool. If you care more about the description file, you are dealing with a workflow.
+
+History
+-------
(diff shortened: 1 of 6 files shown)


denis-yuen (approved):

NatalieEO (approved):

docs/advanced-topics/tools-vs-workflows.rst
```diff
@@ -0,0 +1,117 @@
+Tools vs Workflows: Detailed Comparison
```
coverbeck, line 1: This is the only colon in the sidebar. Suggest removing, which may (or may not) require some rewording

docs/advanced-topics/tools-vs-workflows.rst
```diff
@@ -0,0 +1,117 @@
...
++------------------------+------------------------------------------+-------------------------------------------------+
+
+What about workflows that run in Docker containers?
+---------------------------------------------------
+
+Workflow descriptor files do not require a Docker container. That being said, you can still specify external Docker images
+within the descriptor files, and doing so is generally recommended. Each task (WDL) or step (CWL) can specify a Docker image, which might be totally different from the Docker image used by a previous task/step in the workflow. Because of this, workflows registered on Dockstore that reference Docker image(s) will still follow versioning from GitHub, instead of versioning based on the tags of the Docker image(s).
+
+.. hint::
+    The Cromwell execution engine requires that WDL task specify a Docker image when running on GCP or AWS backends, including Terra. Therefore, if your WDL workflow needs to run on Terra, make sure that each task uses the Docker runtime attribute. For more information see the `Cromwell docs <https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#docker>`_.
```
coverbeck, line 59: "that WDL task specify..." -> "that a WDL task specify..." or "that WDL tasks specify..."

coverbeck, line 59: > WDL task specify a Docker image when running on GCP or AWS backends

Do you mean if you're running Cromwell without Terra? Sounds like... I'm not sure in all cases Cromwell would know it's running on AWS (you're just on EC2 instance), but probably not worth getting into that.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
```diff
@@ -1,98 +1,32 @@
...
-However, our definition for each does not completely align with the language's specification.
-Instead, Dockstore tools are more associated with creating/owning Docker images that are used in conjunction with a descriptor language, and
-Dockstore workflows are more closely tied to the descriptor files themselves. While other descriptor languages, like WDL and Nextflow,
-do not have separate concepts for tools and workflows, we still maintain a distinction between a tool and a workflow for WDL.
+vcf file --> convert to GDS file --> give each variant a unique ID --> a GDS file with unique IDs
 
+In this case, the output of the conversion step is the input of the step that generates unique IDs, and the unique IDs step's output is the final output of the workflow: A GDS file with unique IDs. 
 
+Managing workflows can be complicated, as you may be chaining together many steps, each with different computational requirements. This where workflow languages step in, allowing us to create formal "tasks" which can each have their own computational settings or Docker container, and chaining the output of one task into the input of another task.
```
coverbeck, line 13: > This where workflow

This is where workflow...

coverbeck (approved):

docs/advanced-topics/tools-vs-workflows.rst
aofarrel: Strictly speaking, [Cromwell's docs say this about the docker runtime attribute](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#docker):

> GCP: This attribute is mandatory when submitting tasks to Google Cloud.
> AWS Batch: This attribute is mandatory when submitting tasks to AWS Batch.

Due to how the docs tend to present things as "this is what changes based on the backend," I interpreted this to mean "any AWS backend," but I will be the first to admit I know very little about AWS. What's your interpretation?

coverbeck: My interpretation is that if you use Cromwell with AWS Batch, which will let Cromwell scale up on its own, then you need it.

But if you just ssh into a machine that happens to be hosted on AWS and running Cromwell locally, it doesn't matter.

Same is probably true for GCP...

So, I would either drop the the Cromwell execution engine reference, or if you want to refer to it, maybe talk about in terms of [scaling](https://cromwell.readthedocs.io/en/stable/Scaling/).

aofarrel: Good point -- I clarified it.

PR #167, opened 2022-02-02
Fix readthedocs build

Richard-Hansen:
readthedocs was building from `docs/requirements.txt`, while developers have been building from `./requirements.txt`, this caused some broken builds. To determine a working set of pinned versions, I took the final downloaded versions from a successful build of unpinned requirements, i.e. this build: https://readthedocs.org/projects/dockstore/builds/15967555/


coverbeck (approved):
Just curious, I don't understand why docs/requirements.txt had so many more dependencies that the new ./requirements.txt -- were they unused?

Richard-Hansen:
> Just curious, I don't understand why docs/requirements.txt had so many more dependencies that the new ./requirements.txt -- were they unused?

It was only used for the remote builds, not sure why there are so many more dependencies though... I didn't even realize we had two `requirements.txt` until looking in to this ticket. Perhaps @aofarrel knows more?

aofarrel:
> > Just curious, I don't understand why docs/requirements.txt had so many more dependencies that the new ./requirements.txt -- were they unused?
> 
> It was only used for the remote builds, not sure why there are so many more dependencies though... I didn't even realize we had two `requirements.txt` until looking in to this ticket. Perhaps @aofarrel knows more?

I genuinely have no idea. I had noticed we had two of them but I didn't even know one was for remote builds. I took a bit of a "if in doubt, don't touch it" approach.

PR #183, opened 2022-03-21
Skip all doi.org URLs in linkcheck

aofarrel:
These URLs appear to redirect to a different URL, but linkcheck should be able to handle that. No idea why that's happening but this workaround handles it. I don't like it, but there is already a doi.org URL getting skipped, so maybe this sort of thing has happened before?


denis-yuen (approved):
I don't know either, but this seems reasonable

svonworl (approved):

wshands (approved):

PR #189, opened 2022-04-04
Remove reference to legacy Getting Started with CWL tutorial

aofarrel:
I made last-minute changes to Getting Started with CWL to remove the Docker/Quay stuff in 1.12.0, but forgot to remove this reference to said Docker/Quay stuff in dockstore-tools.rst. This addresses that.


denis-yuen (changes requested):
since this is going into develop, it won't make it into 1.12 documentation

aofarrel:
bah, targeted the wrong branch, thanks for catching that

PR #191, opened 2022-04-04
Proposal: Adding "back to top" links in long documentation pages

aofarrel:
Proof of concept for https://github.com/dockstore/dockstore/issues/4830


aofarrel:
It's not worth dealing with this rebase considering only one page should be changed. Will make a new feature branch.

PR #186, opened 2022-04-01
Update intro page

bethsheets:
- Provided a general intro for tools and workflows
- Defined tool -- it was missing in prior version
- Removed unnecessary information about Dockstore's legacy tool definition
- Removed recommendation bc we may expand tool support for Galaxy in the future

PR #192, opened 2022-04-05
New PR with only my edits cherrypicked for updating intro page

bethsheets:
Prior PR #186 was accidentally based on the wrong branch

### DIFF
--- a/docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
+++ b/docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
@@ -1,27 +1,28 @@
 Introduction to Workflows, Tools, and Services
 ==============================================
 
+Tools and workflows hosted on Dockstore are computational steps described in languages that make it easy for their reuse in different compute environments. Generally, descriptor languages request that a user include a containerized compute environment (such as a Docker image) that packages up all of the code and dependencies needed to complete the computational step.
+
+Not all descriptor languages make a distinction between a tool and a workflow. Most users will register workflows on Dockstore. 
+
 What is a workflow?
 -------------------
 
-In general terms, a workflow is a chain of commands that has several steps linked together, which create some sort of final output. For example, a simple two-step workflow that converts a VCF into a GDS file may conceptually look like this:
+In general terms, a workflow describes a chain of commands in a specific order. For example, a simple two-step workflow that converts a VCF into a GDS file may conceptually look like this:
 
-vcf file --> convert to GDS file --> give each variant a unique ID --> a GDS file with unique IDs
+VCF file --> call tool convert to GDS file --> give each variant a unique ID --> a GDS file with unique IDs
 
 In this case, the output of the conversion step is the input of the step that generates unique IDs, and the unique IDs step's output is the final output of the workflow: A GDS file with unique IDs. 
 
-Managing workflows can be complicated, as you may be chaining together many steps, each with different computational requirements. This is where workflow languages step in, allowing us to create formal "tasks" which can each have their own computational settings or Docker container, and chaining the output of one task into the input of another task.
+Managing workflows can be complex, as you may be chaining together many steps, each with different computational requirements. This is where workflow languages step in, allowing us to create formal "tasks" which can each have their own computational settings or Docker container, and chaining the output of one task into the input of another task.
 
 What is a tool?
 ---------------
-How Dockstore defines a tool is changing, but our aim is to align our tool definition with that of the languages we support.
-For the sake of simplicity, Dockstore only recommends registering a Dockstore tool when your work is a CWL CommandLineTool.
-
-If you want a deeper understanding of the history of Dockstore tools and where we plan on going, read :doc:`Dockstore Tools Overhaul </../advanced-topics/dockstore-tools-overhaul>`.
+A tool represents a single command line tool wrapped in a descriptor language.  Languages that formally describe tools (such as CWL) may chain them together into a workflow.
 
 
 What is a service?
 ------------------
 
-Services are meant to be long running processes, usually web services or interactive applications, that can be launched by a user. 
+Services are meant to be long running processes, usually web services or interactive applications, that can be launched by a user. Dockstore services are currently in beta. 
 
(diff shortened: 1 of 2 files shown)


docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
```diff
@@ -1,27 +1,28 @@
...
 
 What is a tool?
 ---------------
-How Dockstore defines a tool is changing, but our aim is to align our tool definition with that of the languages we support.
-For the sake of simplicity, Dockstore only recommends registering a Dockstore tool when your work is a CWL CommandLineTool.
+A tool represents a single command line tool wrapped in a descriptor language.  Languages that formally describe tools (such as CWL) may chain them together into a workflow.
 
-If you want a deeper understanding of the history of Dockstore tools and where we plan on going, read :doc:`Dockstore Tools Overhaul </../advanced-topics/dockstore-tools-overhaul>`.
 
+Introduction to Services
```
coverbeck, line 24: This comes out looking odd, IMHO:

Screen Shot 2022-04-05 at 9 48 30 AM

Clicking on Introduction to Services in the left TOC takes you to a page whose top title is "Introduction to Workflows and Tools" (the services part is at the bottom, see screenshot).

I'm OK with less highlighting of services, since they are not very useful today. I think you should:

1. Just drop services from this page entirely
2. Drop this from this page, and either create a new page, or move this section to "Services" page
3. Restore them to this page

I lean towards 2.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
```diff
@@ -1,27 +1,28 @@
-Introduction to Workflows, Tools, and Services
+Introduction to Workflows and Tools
 ==============================================
```
coverbeck, line 2: I thought the linter enforced this, but I guess not (I'm confused actually, because I could swear I've seen itin the past), but we typically make the width of this line identical to the width of the line above it.

aofarrel: The linter only enforces if the line is too short, not too long.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
```diff
@@ -1,27 +1,21 @@
...
 
-Managing workflows can be complicated, as you may be chaining together many steps, each with different computational requirements. This is where workflow languages step in, allowing us to create formal "tasks" which can each have their own computational settings or Docker container, and chaining the output of one task into the input of another task.
+Managing workflows can be complex, as you may be chaining together many steps, each with different computational requirements. This is where workflow languages step in, allowing us to create formal "tasks" which can each have their own computational settings or Docker container, and chaining the output of one task into the input of another task.
 
 What is a tool?
 ---------------
-How Dockstore defines a tool is changing, but our aim is to align our tool definition with that of the languages we support.
-For the sake of simplicity, Dockstore only recommends registering a Dockstore tool when your work is a CWL CommandLineTool.
-
-If you want a deeper understanding of the history of Dockstore tools and where we plan on going, read :doc:`Dockstore Tools Overhaul </../advanced-topics/dockstore-tools-overhaul>`.
```
denis-yuen, line 20: Think the link makes sense and already had the right level of suggestion

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
```diff
@@ -1,27 +1,21 @@
...
-
-
-What is a service?
```
aofarrel, line 23: I disagree with the removal of the services section. A user is going to see services mentioned several times in docs relating to .dockstore.yml, but if the section is removed entirely then they need to read the much more lengthy /getting-started/getting-started-with-services.html for context rather then a simple one-sentence introduction.

What Charles pointed out with the headers is indeed an issue, but that can be solved in RST by using the same header type consistently. RST has detected that you started the page with a header that uses = as the underline, so it considers that the top-level header style for this page, and - as the second-level header style. (Note that this is per RST page - our docs are wildly inconsistent across different RST pages. Sometimes ~~~~ is the second-level header style for instance.) So if you introduce services like this:
```
What is a service?
------------------
```
It should be the same type as header as the tool and workflow ones.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
bethsheets: @denis-yuen, do you want me to put this text back?

denis-yuen: yeah, I think so

bethsheets: @denis-yuen In my opinion, this page is aimed directly at new users that don't know what a tool or workflow is. I don't think they need this extra information and it could be more confusing than helpful to them. 

This link is only relevant to returning users. Returning users should be about to find the tool overhaul and how it affects them in our tool registration and advanced topics pages.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
bethsheets: @aofarrel I changed this back to the --- header, but this goes back to your prior request of it being confusing if we don't introduce services at the top alongside workflows/tools intro. Not sure what to do.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
denis-yuen: There's an "if you want", I think they can judge for themselves if they want a deeper understanding. 
It looks like the page could be useful for helping users decide between our various registration methods.

docs/getting-started/intro-to-dockstore-tools-and-workflows.rst
bethsheets: Ash and I agreed on option 2 and moved intro to Services page.

denis-yuen (approved):

aofarrel (approved):
Beth and I conversed on Slack about services and I changed my mind about including services on this page. My concern was not so much interest in services but the fact services are mentioned elsewhere in the docs... but there is a sidebar entry labeled "services" that will handle that case.

coverbeck (approved):

PR #195, opened 2022-04-20
Pin a dependency and clarify .dockstore.yml `filter` is not about Docker tags

aofarrel:
Neither this version nor the old one mention Docker tags at all, but I think users new to Docker and git have a chance of getting the two mixed up. I swapped the order of the words and added "GitHub" right before "tags" in attempt to make it clearer, as well as linked GitHub's own docs on the two. Also included a sentence reminding users that they need to have their workflow/tool/service on Dockstore, just in case a user goes straight here without reading any of our other docs.

### DIFF
--- a/docs/assets/templates/template.rst
+++ b/docs/assets/templates/template.rst
@@ -9,6 +9,7 @@ Not sure if you are working with a workflow, a tool, or a service? :doc:`Check o
 
 Tips and tricks
 ---------------
+* This form of registration is only available to workflows, tools, and services that are hosted on GitHub
 * Make sure your file is saved as ``.dockstore.yml``, not ``dockstore.yml`` or ``.dockstore.yaml``
 * Put the .dockstore.yml file in the top of your repo or inside ``.github/``
 * The first line of a .dockstore.yml file references the version of .dockstore.yml syntax being used, not the version/tag of the workflow/tool/service it describes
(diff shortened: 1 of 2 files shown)


aofarrel:
Screen Shot 2022-04-20 at 4 38 01 PM

...hmmmmmmm.

Screen Shot 2022-04-20 at 4 36 04 PM

aofarrel:
Did a search of the packages that were updated between the CI builds I pulled these from, not seeing any issues about attrs not getting installed.

coverbeck (approved):
A little confused why requirements.txt is needed all of the sudden. Is it because we're not pinning all our requirements.

Going to defer to @Richard-Hansen or any other Python people he wants to bring in, but text changes look good.

coverbeck (changes requested):
I built this locally and the links don't look right, I think you've done markdown

Screen Shot 2022-04-21 at 4 47 14 PM

Richard-Hansen (approved):
After link change suggested by Charles.

aofarrel:
> > A little confused why requirements.txt is needed all of the sudden. Is it because we're not pinning all our requirements.
> > Going to defer to @Richard-Hansen or any other Python people he wants to bring in, but text changes look good.
> 
> I'm not seeing why our dependencies changed requirements. Our versions are pinned and the git repo hasn't changed in quite a while. Perhaps some deep-down transitive dependency? Regardless, probably nothing to do on our end.

Our versions are pinned, but their own dependencies presumably are not themselves pinned. For example, an older build installed mdurl-0.1.0 while a newer one installed mdurl-0.1.1 so I'm lead to believe that some dependency of a dependency is not installing a dependency of that dependency's dependency. Given the error is in myst_parser it is probably one of its dependencies but I checked the packages that have been updated between the two builds screenshot above, and didn't find the culprit, nor did I find people reporting issues... so I'm just pinning attrs in our own requirements.txt whilst shrugging.

coverbeck:
> Our versions are pinned, but their own dependencies presumably are not themselves pinned. For example, an older build installed mdurl-0.1.0 while a newer one installed mdurl-0.1.1 so I'm lead to believe that some dependency of a dependency is not installing a dependency of that dependency's dependency. Given the error is in myst_parser it is probably one of its dependencies but I checked the packages that have been updated between the two builds screenshot above, and didn't find the culprit, nor did I find people reporting issues... so I'm just pinning attrs in our own requirements.txt whilst shrugging.

Speaking in general terms, it's odd to me that all transitive dependencies are not captured in the requirements.txt file. In the npm/JavaScript world they are.

If transitive dependencies aren't captured, then the requirements.txt is going to be prone to breakage.

coverbeck (approved):

PR #196, opened 2022-04-26
Shove 1.1 template + square renderer into attic

aofarrel:
As part of reorganizing our docs, it makes sense to put a few things into the corner. This is not just for outdated docs, it's for "meta" pages such as the HTML document I use to create square icons on the homepage. 

By putting these things into their own folder and adding an exclusion to conf.py, these files do not get compiled at all, saving a little bit of time.


aofarrel:
In the future I'll move more things into this folder. Starting small for now in this PR.

david4096 (approved):
Looks good

PR #198, opened 2022-04-28
Add dates and context to news pages

aofarrel:
https://github.com/dockstore/dockstore/issues/4878

Seeking feedback on what to do about 2017-05-05-Upcoming-Features.rst, which seems to have two news articles in it but it's not clear which truly comes first (the May 5th date is a guess based on the file name, but a typo could have put it at April 5th), and if we should shove these into the _attic (see https://github.com/dockstore/dockstore-documentation/pull/196).


denis-yuen (approved):
Think the changelog comes first, looks like we had a pretty good idea of what would come next after that release so we gave people a preview.

bvizzier (approved):
Looks good.  I like the changes.

PR #199, opened 2022-05-02
Replace wonky link in docker-alternatives.rst

aofarrel:
We have a link to Docker's blog r/e rootless Docker. Recently, it has begun occasionally but not always failing during `make linkcheck`. As of my making this PR, builds are passing on CircleCI but my browser is still throwing gateway errors on that URL.

It just so happens the author of the blog mirrored it on Medium, so this is an easy fix. Will add reviewers + look to merge this if CI begins failing again. Might just be a temporary hiccup.

### DIFF
--- a/docs/advanced-topics/docker-alternatives.rst
+++ b/docs/advanced-topics/docker-alternatives.rst
@@ -7,7 +7,7 @@ Docker Alternatives
 In some situations using Docker may be impractical because it requires all users to have root access.
 Several alternatives have been developed to make it possible to run rootless containers, including
 `Singularity <https://sylabs.io/docs/>`_ and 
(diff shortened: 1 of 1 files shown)


aofarrel:
Turns out that the original post is still online, it's just a different URL. engineering.docker.com redirects to docker.com/blog/ and with a bit of experimentation I managed to derive https://www.docker.com/blog/experimenting-with-rootless-docker/.

aofarrel:
Updated with the docker.com version as it is a little more "official" than a medium post.

denis-yuen (approved):

kathy-t (approved):

aofarrel:
This is causing issues with linkcheck in #200 and #201 so this will be pulled into those PRs too.

PR #200, opened 2022-05-16
Implement dockstore dictionary

aofarrel:
Implements a Dockstore dictionary. Did my best to make all of the entries accurate, **but please check them for accuracy and add suggestions on what needs to be added or changed.**

* File is generated by a Python program I wrote, and should be updated by only that program 
  * I will release it pypi once I heavily reorganize [the repo it came from](https://github.com/aofarrel/Dockstore-Dictionary) and come up with a snappy name for it; current working title is GreatGloss but I'm open to suggestions
  * Once it's in a release state, I'll update this PR/open a new PR to add the Python source file and put it in _attic.
* This PR does not include any other links to documentation to keep it manageable. In other words this dictionary does not get linked anywhere except on the sidebar.


aofarrel:
usegalaxy is currently down; ignore the output of `make linkcheck` for now. Will update this PR if things remain finicky.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
+.. This file was created using GreatGloss. It is highly recommended to update the
```
coverbeck, line 1: It shouldn't be `highly recommended`, it should be marked as a generated file with "DO NOT UPDATE" as a comment, or it should always be updated manually.

coverbeck (commented):
I left some comments, but it's a lot of text to go through. And very impressive!

But some broader thoughts:

1. We should not merge this until the tooling is in place to update it by anybody on the team. What if there's a hideous typo? If we fix it directly in here, we risk losing the fix if it gets regenerated.
2. Python has generally not been a skill needed to work on the Dockstore team; many of us the team don't know it (many do, to be fair). There are exceptions, e.g., this repo itself is built with Python, and we may have to do down that road for Snakemake support, but I'm a tad wary of introducing more 

This is not a judgement on Python, but the team has to know Java, TypeScript along with the particular libraries we use for each language, AWS, CloudFormation, etc. I'd rather not add one more library to maintain. Again what if there is a bug in the generator and you're on vacation? How do I push an update to PyPI? I'm sure there's a way, but we'd have to figure it out.

But I may be overreacting, so curious what others think.

aofarrel:
Regarding the Python codebase: Although dockstore.org deals with TS, docs.dockstore.org has always been Pythonic and required Python-related maintenance (ex: The dependency issues I ran into and solved in https://github.com/dockstore/dockstore-documentation/pull/195 and the handful of times I've had to edit conf.py). I would say that the amount of Python I am adding here is less complex than the Python we are already using to generate the docs. The generation of this file can also be plugged into Sphinx's generation and/or our Makefile, although my current priority is to package the Python generator code so it can be imported first (I could swap priorities if you prefer).

Regarding team speciality: I was thinking the generator to be a TLP sort of thing, while Dockstore only has to deal with the [barely-Python input](https://github.com/aofarrel/Dockstore-Dictionary/blob/main/entries.py) and the RST output. TLP have written Python tools for specific tasks before and maintained them after a team member has left such as https://github.com/DataBiosphere/herzog

> How do I push an update to PyPI?

Strictly speaking you wouldn't need to. In a "this needs to be edited or else our docs will not compile in time for our next release" scenario, the .py file that contains the generating module can just be put in the same directory (or even the same file) as the .py file containing the entry definitions. In a non-emergency "Ash is sailing around the world and we just want to update the package in case anyone else is using it" scenario, TLP we be given the required permissions to update.

denis-yuen:
> This is not a judgement on Python, but the team has to know Java, TypeScript along with the particular libraries we use for each language, AWS, CloudFormation, etc. I'd rather not add one more library to maintain. Again what if there is a bug in the generator and you're on vacation? How do I push an update to PyPI? I'm sure there's a way, but we'd have to figure it out.

> It shouldn't be `highly recommended`, it should be marked as a generated file with "DO NOT UPDATE" as a comment, or it should always be updated manually.

> Regarding the Python codebase: Although dockstore.org deals with TS, docs.dockstore.org has always been Pythonic and required Python-related maintenance (ex: The dependency issues I ran into and solved in https://github.com/dockstore/dockstore-documentation/pull/195 and the handful of times I've had to edit conf.py). I would say that the amount of Python I am adding here is less complex than the Python we are already using to generate the docs. The generation of this file can also be plugged into Sphinx's generation and/or our Makefile, although my current priority is to package the Python generator code so it can be imported first (I could swap priorities if you prefer).

Threading the needle a bit on this. Hiring *is* difficult with the variety of languages and libraries we use. This includes the workflow languages and the supporting libraries/engines themselves. That said, we do use a variety of CI platforms and technologies. But this isn't really hooked up as such. 

Proposing that we do so: so this should be hooked up to github actions so it runs on every build and we get advanced notice of breakage, that it regenerate the dictionary on every build (so to make it even more obvious that changes to the dictionary will be wiped out), and that the dictionary source (`entries.py`) be checked in here. Might not even need to check-in the generated source (`docs/dictionary.rst`)

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict AWS:
+
+AWS
+---
+*abbreviation for* Amazon Web Services  
+
+	A provider of cloud services, most notably cloud computing and cloud storage, available on-demand and hosted by Amazon. :ref:`dict Seven Bridges` is an example of a system that is powered by AWS, and can launch workflows on :ref:`dict EC2` instances.
```
denis-yuen, line 132: A little off, describing a well known system like AWS in terms of Seven Bridges which is not.

Maybe something like "Netflix and AirBnb are examples of systems powered by AWS and this includes some systems in bioinformatics like Seven Bridges"
https://aws.amazon.com/solutions/case-studies/innovators/

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict collections:
+
+collections
+-----------
+	A group of workflows or tools on Dockstore associated with a particular :ref:`dict organization`.
```
denis-yuen, line 241: But why is it associated with a particular organization (i.e. we should probably describe our stuff a little better)

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict Common Workflow Language:
+
+Common Workflow Language
+------------------------
+	A workflow language that describes how to run command-line tools. CWL is based on Java and can use Java commands within its own commands. :ref:`dict WDL` and CWL are relatively similar in principle, and code written in one language can often be translated into the other with some workarounds, but they are two different standards and each have unique features.
```
denis-yuen, line 252: CWL doesn't have much to do with Java, should probably just take that out. It can use CLI programs from slightly above.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict DOI:
+
+DOI
+---
+*abbreviation for* Digital Object Identifier  
+
+	An identifier that provides a long-lasting link to some sort of :ref:`dict immutable` digital object. On Dockstore, you can use Zenodo to mint a DOI of your workflows and tools to increase reproducibility.
```
denis-yuen, line 406: Should elaborate that DOIs are useful for citing programs in publications. 
Software should be cited like all other research products, etc 
https://cite.research-software.org/

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict GCP:
+
+GCP
+---
+*abbreviation for* Google Cloud Platform  
+
+	A backend used for cloud computing and cloud storage hosted by Google. :ref:`dict Terra` is an example of a system that runs on a GCP backend. When running workflows on these backends, make sure to account for the storage needed for your workflow, as GCP compute backends do not automatically scale their storage size at runtime. GCP backends allow you to make use of Google's :ref:`dict preemptible` feature, which may reduce the cost of running workflows.
```
denis-yuen, line 553: Same deal as with AWS, pick a more well known example than Terra and then lead to it

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict labels:
+
+labels
+------
+	On Dockstore, we use labels to "tag" Dockstore entries with information about them. You can add labels to a Dockstore :ref:`dict entry` page that you have edit access to. Labels cannot contain spaces.
```
denis-yuen, line 699: Should be more specific about "you" here which is a workflow developer, distinct from categories or collections above.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. _dict ORCID:
+
+ORCID
+-----
+pronounced "or-kid", rhymes with kid  
+
+*abbreviation for* Open Researcher and Contributor ID  
+
+	ID used to identify researchers and their work in a way that doesn't solely rely on names.
```
denis-yuen, line 886: I think unique is key here

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+Further reading: `<https://info.orcid.org/what-is-orcid/>`_  
+
+.. updated 2022-05-16  
+
+
+
+.. _dict organization:
+
+organization
+------------
```
denis-yuen, line 897: Users also should have the authority to speak for the institute, grant, etc. in a technical manner

bethsheets (changes requested):
This was a lot of work! I had some suggestions on definitions. 

I think Dockstore team should lead on helping you resolve the question around using a python tool.

aofarrel:
> Proposing that we do so: so this should be hooked up to github actions so it runs on every build and we get advanced notice of breakage, that it regenerate the dictionary on every build (so to make it even more obvious that changes to the dictionary will be wiped out), and that the dictionary source (`entries.py`) be checked in here. Might not even need to check-in the generated source (`docs/dictionary.rst`)

Although I didn't commit it to this repository to keep this PR small, I already have [a pre-commit hook](https://github.com/aofarrel/Dockstore-Dictionary/blob/main/hooks/pre-commit) and [a testing setup](https://github.com/aofarrel/Dockstore-Dictionary/blob/main/Makefile) on the repository where entries.py lives that regenerates the dictionary from entries.py every time I commit. Originally I was going to open a new PR with that hook/stuff from the makefile, but I think doing it on build like you suggested is a better approach.

In any case, I'll open new PR targeting this PR to address these concerns.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict Jupyter:
+
+Jupyter
+-------
+pronounced "Jupiter" like the planet  
+
+	A project focused on developing "notebooks" for programming languages, most famously Python due to it starting as a splinter of iPython in the early 2010s, but including other languages as well such as R. Jupyter notebooks allow for blocks of code to be nestled between markdown text, allowing for easy documentation of the code blocks and reproducibility of analysis.
```
avani-k, line 661: Nitpicking here. The first sentence here seemed a bit bulky to me. Suggested reframing:

A project focused on developing "notebooks" for programming languages, most famously for Python due to it starting as a splinter of iPython in the early 2010s. But Jupyter notebooks are used for other languages as well, such as R.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict WDL:
+
+WDL
+---
+pronounced "widdle", rhymes with little
```
avani-k, line 1037: Nitpicking. Maybe riddle instead of little?

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict YAML:
+
+YAML
+----
+*abbreviation for* YAML Ain't Markup Language  
+
+	Human-readable data-serialization language. Commonly used for configuration files.
```
avani-k, line 1092: Nitpicking!

pronounced "yah-ml", rhymes with camel

avani-k (approved):
Few minor comments, mostly nitpicking! Great job Ash!

coverbeck:
>I would say that the amount of Python I am adding here is less complex than the Python we are already using to generate the docs. The generation of this file can also be plugged into Sphinx's generation and/or our Makefile

Good points.

> Strictly speaking you wouldn't need to. In a "this needs to be edited or else our docs will not compile in time for our next release" scenario, the .py file that contains the generating module can just be put in the same directory

Following up on this echoing Denis' suggestion, could you just put the Python code in this repo from the beginning and make it part of `make`? As you suggest above?

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+	* :ref:`dict Workflow Execution Service`
+	* :ref:`dict YAML`
+
+.. _dict .dockstore.yml:
+
+.dockstore.yml
+--------------
+	This file is part of :ref:`dict GitHub App registration`. It indexes workflows or tools within a repository, including their optional test parameter files, and the author(s) of said workflows or tools.  
+
+Further reading: `<https://docs.dockstore.org/en/stable/assets/templates/template.html>`_
```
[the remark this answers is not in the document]
aofarrel: Implemented a fix, but I'm going to leave this unresolved as I'll want to double check once it is published.

docs/dictionary.rst
aofarrel: How about: "A provider of cloud services, most notably cloud computing and cloud storage, available on-demand and hosted by Amazon. Netflix and AirBnB are examples of a system that is powered by AWS. Bioinformatics systems such as [Seven Bridges] can levarage AWS by launching workflows on [EC2] instances."

docs/dictionary.rst
aofarrel: Agreed, which is why I was hoping to get some input on a few of these entries. I get that collections tend to be all for a particular purpose, but the difference between that and a category can get fuzzy - if the Broad organization has a collection for GATK workflows, the meaningful difference between that and a category of GATK workflows is just that the collection implies that those workflows are maintained by the Broad while the category does not imply that. But as soon as I say that in a definition, it implies that collections are always maintained by their organization, ie it puts a burden on organizations. How should I approach this?

docs/dictionary.rst
aofarrel: Does this apply to the first sentence or the second? ie are users within an organization have that authority, or only users who are applying to have an organization created?

docs/dictionary.rst
aofarrel: I suppose I was incorrect when I was talking about Java, but I think the ability to use Javascript all over the place is an important feature. Notably, there is no equivalent to this in WDL. 

https://www.commonwl.org/user_guide/13-expressions/index.html

docs/dictionary.rst
denis-yuen: That's more thought than I had put into it. I was thinking something like "A collection is a group of Dockstore workflows that an organization found useful, created themselves, or found interesting. See their description to see why the organization collected workflows/entries in a collection"

docs/dictionary.rst
denis-yuen: Oh, Javascript expressions. 
Yes, I think it could be worthwhile to point out Javascript expressions.

docs/dictionary.rst
denis-yuen: I think the latter. We don't really check on the former and there are different roles anyway, some of which clearly have lesser authority.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+	An identifier that provides a long-lasting link to some sort of :ref:`dict immutable` digital object. On Dockstore, you can use Zenodo to mint a DOI of your workflows and tools to increase reproducibility.  
+
+
+.. updated 2022-05-16  
+
+
+
+.. _dict DS-I Africa:
+
+DS-I Africa
```
coverbeck, line 415: TIL :) I hadn't heard of this one before.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict EC2:
+
+EC2
+---
+*abbreviation for* Elastic Compute Cloud  
+
+	The cloud computing side of :ref:`dict AWS`. When running workflows on these backends, disk size will scale with your workflow requirements automatically. EC2 instances allow you to make use of Amazon's :ref:`dict spot instance` feature, which may reduce the cost of running workflows.
```
coverbeck, line 433: 1. I don't know about disk sizes scaling automatically. I don't think it's an EC2 feature, but I could be wrong.
2. EC2 instances can be spot instances, I wouldn't say they allow you to make use of them.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict egress:
+
+egress
+------
+pronounced "ee-gress", rhymes with aggress
```
coverbeck, line 445: Interesting, I'd never heard the word `aggress` and had to look it up, although I could sort of derive it from `agression`.

Also, the stress in `aggress` is different than `egress` (last syllable vs. first syllable), is that still considered a rhyme? I guess so.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
...flow`, or :ref:`dict Galaxy workflow` syntax. An entry can optionally include a :ref:`dict parameter file` that links to open-access test data. A single entry will include all versions of the tool or workflow that has been registered, with that versioning being based upon the versioning and branches of the source-control repository the descriptor file is hosted on (with the exception of a :ref:`dict legacy tool`, which have versioning based upon their Docker image tags), and any version can be pinned as the default. Entries can be added to :ref:`dict collections` associated with a particular :ref:`dict organization`, or added to :ref:`dict categories` so they can be grouped with other entries that have a similar scientific purpose. Entries may also have :ref:`dict labels` attached to them to help them be found via Dockstore's :ref:`dict faceted search` feature. If the entry is registered using the :ref:`dict Dockstore GitHub App`, then the entry will stay in sync automatically with the source-control repository. Additionally, if an entry is a valid :ref:`dict workflow`, any user can use our :ref:`dict launch with` feature to import the workflow to one of our cloud compute partners.
```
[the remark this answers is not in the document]
coverbeck: I would tend to keep this short and just describe it as shorthand for a tool or workflow. But I'm biased because I don't think we should even surface entry at all, although the cat's out of the bag on that one.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict NIH:
+
+NIH
+---
+*abbreviation for* National Institute of Health
```
coverbeck, line 854: Institute**s**

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict parameter file:
+
+parameter file
+--------------
+	A :ref:`dict JSON` or :ref:`dict YAML` file that describes the inputs to a workflow. This usually includes internal links, or links to data in a Google or S3 bucket.
```
coverbeck, line 910: What is an internal link?

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict preemptible:
+
+preemptible
+-----------
+	A type of :ref:`dict GCP` :ref:`dict VM` which may have its running jobs interrupted at any given time, and will be shut down if running for more than 24 hours. A preemptible machine is significantly cheaper than a standard VM, at the cost of possibly stopping before your computational work is finish. You can use preemptible machines when running workflows on GCP backends to save on compute costs.
```
coverbeck, line 921: finish**ed**

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. _dict workflow:
+
+workflow
+--------
+	A command line program wrapped in a descriptor language, which usually has multiple steps. In :ref:`dict CWL`, a workflow is usually made up of multiple tools. Other languages consider a workflow to be the basic unit.  
+
+see also :ref:`dict tool`  
+
+Further reading: `<https://docs.dockstore.org/en/stable/getting-started/intro-to-dockstore-tools-and-workflows.html>`_
```
coverbeck, line 1054: Relative link

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict Workflow Execution Service:
+
+Workflow Execution Service
+--------------------------
+	A standardized API developed by :ref:`dict GA4GH` for describing a standard programmatic way to run and manage workflows.
```
coverbeck, line 1078: Optionally mention Dockstore CLI supports running workflows via WES.

coverbeck (commented):
After going through this a little bit more thoroughly, I'm even more  impressed with all the work you put into this.

docs/dictionary.rst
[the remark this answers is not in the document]
aofarrel: I was using Seven Bridges in order to contrast GCP/Terra, but you're right, this might lead to some confusion even though the SB entry does note it can use GCP too. 

In response to other feedback, the AWS entry currently sits at:

> A provider of cloud services, most notably cloud computing and cloud storage, available on-demand and hosted by Amazon. Netflix and AirBnB are examples of a system that is powered by AWS. Some bioinformatics systems such as [Seven Bridges] can leverage AWS by launching workflows on [EC2] instances.

...which no longer implies SB relies upon AWS, but rather *can* use it. Do you think that is sufficient or should it be excised completely? I think a bioinformatics example is needed, but I also think WES may be a little too obscure to be our example... but maybe there is another option besides AWS, or maybe WES is the way to go here.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict egress:
+
+egress
+------
+pronounced "ee-gress", rhymes with aggress  
+
+	The action of leaving a place. In the context of :ref:`dict cloud computing`, an egress charge is a fee charged for downloading a file. Sometimes, the person hosting the file is charged for data egress. Other times, the person downloading the file is charged.
```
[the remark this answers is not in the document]
aofarrel: Since "egress" shows up in so many forms (data egress, egress charge, egress fee, on its own such as "we have to pay a lot for egress"), I'm leaning towards keeping the definition's title as egress. That being said your suggested definition is more technically accurate/overall helpful than what I wrote. How about this?

> The action of leaving a place. In the context of [cloud computing], data egress refers to data being moved from one location to another, such as from the cloud to a local machine, between cloud providers, and between locations of a single cloud provider. Data egress often results in the charge of fees (usually called egress charges). Data egress can be one of the most expensive cloud costs incurred. Sometimes, the person hosting the file is charged for data egress. Other times, the person downloading the file is charged (such as when downloading files from a Google bucket that has the requester-pays option enabled).

docs/dictionary.rst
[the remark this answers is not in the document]
aofarrel: I might be misunderstanding what Dockstore envisions an entry to be, but I don't think it's correct to call an entry a shorthand for a tool/workflow any more than it would be say that a GitHub repository is a shorthand for a computer program. An entry can contain a tool/workflow, but it's more than just that.

That said, I am open to keeping this definition in short in favor of https://github.com/dockstore/dockstore/issues/4908 instead.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict Terra:
+
+Terra
+-----
+	A cloud-based workflow execution platform developed by the Broad Institute. Terra supports the execution of :ref:`dict WDL` workflows, Jupyter/R notebooks, and integrated apps such as a DICOM-file viewer. The computational backend of a Terra workspace is based upon Google, allowing Google-specific features such as :ref:`dict preemptible` machines to be used in workflows. Dockstore supports directly importing :ref:`dict WDL` workflows into a Terra workspace. Terra is part of the :ref:`dict BioData Catalyst`, :ref:`dict AnVIL Project`, and :ref:`dict eLwazi` grants.
```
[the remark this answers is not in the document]
aofarrel: Fixed, also changed Seven Bridges' entry to say consortium instead of grant.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+.. _dict Workflow Description Language:
+
+Workflow Description Language
+-----------------------------
+	A workflow language managed by the Open WDL Project that is designed to describe command-line tools. Usually written as :ref:`dict WDL`. WDL and :ref:`dict CWL` are relatively similar in principle, and code written in one language can often be translated into the other with some workarounds, but they are two different standards and each have unique features.  
+
+see also :ref:`dict WDL`
```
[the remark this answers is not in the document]
aofarrel: Instead of WDL?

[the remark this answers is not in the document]
aofarrel: The way our documentation will render in HTML will not make it clear where one link begins and another ends, so I'm very hesitant to have more than one link under `see also`

docs/dictionary.rst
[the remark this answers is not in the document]
bethsheets: This new phrasing sounds good to me!

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict immutable:
+
+immutable
+---------
+	Unchanging, unable to be modified. Immutability implies that an object cannot be updated.
```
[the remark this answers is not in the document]
bethsheets: Ah ok, you can ignore my suggestion

docs/dictionary.rst
aofarrel: I may know a lot of rare words, but I don't know how to properly scramble eggs, so it's a tradeoff I suppose...

docs/dictionary.rst
aofarrel: EC2 can autoscale and Amazon markets this feature heavily, but I might be slightly incorrect on the disk size. This is what [Cromwell says about disk size on AWS](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/):

>For the AWS Batch backend, the disk volume is managed by AWS EBS with autoscaling capabilities. As such, the Disk _[sic]_ size and disk type will be ignored. If provided, the mount point will be verified at runtime.

I was conflating EBS with EC2, although strictly speaking EBS works alongside EC2. I don't know for sure if AWS Batch is the only way to run workflows on EC2... how would you word this?


Regarding spot instances, if you choose an AWS backend for your Seven Bridges workspace, you can select spot instances. I've used them myself, and there's docs here: https://sb-biodatacatalyst.readme.io/docs/about-spot-instances
Looks like TOIL supports it too (search [this page](https://toil.readthedocs.io/en/3.15.0/running/cloud/amazon.html) for "spot instance")
But unlike [preemptibles](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#preemptible), this isn't something you put into a runtime attribute in CWL/WDL.

docs/dictionary.rst
aofarrel: I meant internal to the repo. How's this instead?
> A [JSON] or [YAML] file that describes the inputs to a workflow. This usually includes internal links to files within the same source-control repository, or links to data in a Google or S3 bucket.

docs/dictionary.rst
```diff
@@ -0,0 +1,1101 @@
...
+
+.. updated 2022-05-16  
+
+
+
+.. _dict Cromwell:
+
+Cromwell
+--------
+	An open-source :ref:`dict WDL` executor managed by the Broad Institute. Cromwell is the default executor for the :ref:`dict Dockstore CLI` and is the executor used by :ref:`dict Terra`.
```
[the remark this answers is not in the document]
aofarrel: I'm of the belief that calling Cromwell a "workflow management system" isn't as helpful/specific as calling it an executor. It manages the execution of workflows, I guess, but doesn't manage workflows in the sense of source control.

aofarrel:
At Ben's request, entries for AGC and Amazon Genomics CLI have also been added.

docs/dictionary.rst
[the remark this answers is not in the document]
coverbeck: I now better understand our differences on this.

In our Java code for Dockstore we have:

```
abstract class Entry {
...
}

public class Tool extends Entry {
...
}
public class Workflow extends Entry {
...
}
```

So, when I say a workflow is an entry, it literally is that in our class hierarchy behind the scenes. And it's why we use the word entry to refer to a tool or a workflow. It's why `My Entries` on the home page searches your tools and workflows.

But I think you've interpreted it in a different way, where the workflow is the actual executing code, and all the other stuff associated with it is the entry.

While we don't have to echo the class hierarchy in our public facing view, we have so far. As you suggest,  we can revisit with dockstore/dockstore#4908.

docs/dictionary.rst
```diff
@@ -0,0 +1,994 @@
...
+*abbreviation for* :ref:`dict Amazon Genomics CLI`  
+
+
+
+
+.. _dict Amazon Genomics CLI:
+
+Amazon Genomics CLI
+-------------------
+    A :ref:`dict CLI`-based tool that supports launching bioinformatics-related workflows on :ref:`dict AWS` cloud infrastructure. Supports the :ref:`dict WES` standard, and can communicate with the :ref:`dict Dockstore CLI`.
```
coverbeck, line 113: Nit: The Dockstore CLI communicates with AGC. You could argue that communication is two-way, but it is the CLI that always initiates the communication/connection.

docs/dictionary.rst
```diff
@@ -0,0 +1,994 @@
...
+
+Further reading: `<https://commonfund.nih.gov/kidsfirst/highlights>`_  
+
+
+
+.. _dict labels:
+
+labels
+------
+    On Dockstore, we use labels to "tag" Dockstore entries with information about them. Workflow or tool developers can add labels to a Dockstore :ref:`dict entry` page that they have edit access to. Labels cannot contain spaces.
```
coverbeck, line 632: Instead of "Labels cannot contain spaces", I would mention that labels show up on the search page.

docs/dictionary.rst
```diff
@@ -0,0 +1,994 @@
...
+
+
+
+.. _dict NCPI:
+
+NCPI
+----
+*abbreviation for* NIH Cloud Platform Interoperability  
+
+    An effort to connect five :ref:`dict NIH` cloud projects and ensure they are interoperable. The five projects covered under this are the :ref:`dict AnVIL Project`, :ref:`dict BioData Catalyst`, Cancer Research Data Commons, Kids First, and the National Center for Biotechnology Information.
```
coverbeck, line 706: You could add a `:ref` for "Kids First"

docs/dictionary.rst
aofarrel: To avoid getting too deep in the details of this, I think it's better to leave it as is unless there truly is no communication between the two.

docs/dictionary.rst
aofarrel: I've decided to nix the bit about automatically scaling, since I'm not 100% sure on that.

docs/dictionary.rst
coverbeck: It doesn't communicate with the Dockstore CLI. You should drop it or reverse it.

docs/dictionary.rst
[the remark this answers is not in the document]
aofarrel: On second thoughts, this is a good idea. It also encouraged me to [make changes to glossarpy](https://github.com/aofarrel/glossarpy/commit/f39e04393b38a182db3dacca07d7ea62877f9074) that would support this. So now that's simpler too. Thanks for bringing this up!

docs/dictionary.rst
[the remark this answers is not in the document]
bethsheets: LGTM

docs/dictionary.rst
aofarrel: sorted this one out in Slack, marking as resolved

docs/dictionary.rst
coverbeck: I would say relative or local links in that case. But:

1. There are also non-file links in there
2. You typically wouldn't want local/relative links if you were going to run it on something like Terra.

So I would suggest just dropping the internal/relative/local links section, and have something like "such as links to cloud data and runtime parameters".

docs/dictionary.rst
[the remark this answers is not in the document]
coverbeck: I'd prefer execution engine over executor, but I can live with executor.

Still think you should say "Cromwell is the default WDL executor (or execution engine, if you choose that) for the Dockstore CLI.

docs/dictionary.rst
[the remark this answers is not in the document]
coverbeck: I still think entry should be defined to as shorthand for tool or workflow. If we want to broaden the definition, we should discuss later as you suggested. As far as I know, we don't refer to entry any other way.

denis-yuen (commented):
was here, request re-review when hooked up to CI, looking good so far!

aofarrel:
Sorry about the goofy shuffling of commits, I messed up some cherrypicking in attempt to fix the CI. Doesn't affect the actual content of the dictionary, just links that had died elsewhere.

bethsheets (approved):
Looks great! A monumental effort and I know this will be valuable to lots of users (and incoming CGP folks). Thanks so much for creating!

PR #204, opened 2022-05-25
Clarify our usage of Docker and its alternatives, small adjustments to Getting Started series

aofarrel:
GS = getting started

* Adjust GS to Docker to be a little less sales-y
* Adjust GS to Docker to have a more helpful intro
* Adjust some notes on being able to actually run Docker
* Improve intro to our GS series


aofarrel:
Added:
* A wikipedia link to explain what a UID is
* A link to install Docker Desktop, which is now recommended over Docker Engine for most users (didn't delete Docker Engine link as Docker makes it very clear at the top of the webpage that Docker Desktop is preferred)
* Galaxy to list of languages we support
* A few clarifications

Fixed:
* Removed a stray line that implies we only support CWL by saying CWL is next, right above a section that lists non-CWL languages too 
* Some minor grammar issues
* Non-alignment with the style guide

I will not fix the "newlines every X characters" issues until this gets reviewed so reviewers don't need to read past a bunch of changes that are just spacing changes.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -122,13 +125,13 @@ This switches to the ``root`` user to perform software installs. It
...
 
-**This is why Docker is so powerful.** On HPC systems the above process
+This is one of the main advantages of Docker. On some systems the above process
 might take days or weeks of working with a sys admin to install
-dependencies on all compute nodes. Here I can control and install
-whatever I like inside my Docker image - correctly configuring the
-environment for my tool and avoiding the time to set up these
-dependencies in the places I want to run. This greatly simplifies the
-install process for other users that you share your tool with as well.
+dependencies. Instead of installing dependencies on every machine someone might want to run a workflow/tool on, we can instead control and install
```
avani-k, line 130: Suggestion: 

Instead of installing dependencies on every machine one might want to run a workflow/tool on, we can instead install whatever we like inside a Docker image. In this way, we can correctly configure the
environment for a tool and avoid the time to set up these
dependencies in all the systems we might want to run the workflow/tool on. This greatly simplifies the
install process for other users.

avani-k: I split one long sentence into two. And made some subtle changes.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -156,18 +159,17 @@ ownership easier and the default command for this Docker image is set to
...
+for an in-depth look at writing Dockerfiles.
 
 Read more on the development process at
 `https://docs.docker.com <https://docs.docker.com/>`__. For information
 on building your Docker image on Quay.io we recommend their
 `tutorial <https://quay.io/tutorial/>`__.
 
-Building Docker Images
-~~~~~~~~~~~~~~~~~~~~~~
+Build a Docker Images
```
avani-k, line 169: No s at the end of Images

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -193,7 +194,7 @@ should see something like:
 
     Successfully built 01a7ccf55063
 
-Check that the tool is now in your local Docker image cache:
+It might have a different name than ``01a7ccf55063`` but it should a success regardless. Check that the tool is now in your local Docker image cache:
```
avani-k, line 197: Missing verb:
... but it should **BE** a success regardless.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -239,7 +239,7 @@ from the 1000 Genomes project:
     $> /usr/local/bin/bamstats 4 NA12878.chrom20.ILLUMINA.bwa.CEU.low_coverage.20121211.bam
 
 What's really going on here? The ``bamstats`` command above is a simple
-script I wrote to make it easier to call BAMStats. This is what I used
+script we wrote to make it easier to call BAMStats. This is what we used
```
avani-k, line 242: Sentence reads a bit off:

This is what we used the ``COPY`` command to move into the Docker image via the Dockerfile.

docs/getting-started/getting-started-with-docker.rst
aofarrel: I tried to iterate on what you wrote a little bit for greater clarity. How about this?

> This is one of the main advantages of Docker. On some systems the above process might take days or weeks of working with a sysadmin to install dependencies. Instead of installing dependencies on every machine someone might want to run a workflow/tool on, we can create a Docker image that wraps that workflow/tool. With Docker, we only need to correctly configure the environment and dependencies for a program once, and once that Docker image is set up, we avoid repeating that process. All we need to do is transfer that Docker image to other machines, and they will be able to run what's inside it. This greatly simplifies the process for other users.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -1,36 +1,39 @@
...
 Getting Started with Docker
 ===========================
 
 Tutorial Goals
 --------------
+This is the first part of our "Getting Started" tutorial series, where we will walk you through the creation of a tool called BAMstats and publish it onto Dockstore. This particular page, "Getting Started with Docker," will go through the basics of Docker and have you develop a Dockerfile for the BAMstats tool.
 
 -  Learn about Docker
--  Create a Docker image for a real tool
+-  Create a Docker image for a real tool using a Dockerfile
```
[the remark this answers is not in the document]
aofarrel: I think it's clearer if I split it into two sentences...

```
-  Create a Dockerfile
-  Use the Dockerfile you created to make a Docker image for a real bioinformatics tool
```

docs/getting-started/getting-started-with-docker.rst
aofarrel: Rewrote this heavily for clarity; turns out it references an earlier script.

avani-k (approved):

docs/getting-started/getting-started-with-cwl.rst
```diff
@@ -1,4 +1,4 @@
-.. note:: This tutorial assumes basic familarity with Docker as its example involves a Docker image. You may wish to start with :doc:`Getting Started With Docker </getting-started/getting-started-with-docker>` if you are not familiar with it.
+.. note:: This tutorial assumes basic familarity with Docker as its example involves a Docker image, and have access to a system that can run Docker. You may wish to start with :doc:`Getting Started With Docker </getting-started/getting-started-with-docker>` if you are not familiar with it.
```
svonworl, line 1: This sentence doesn't read quite right, there might be a grammar error.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -1,36 +1,42 @@
...
-.. note:: This is the first part of a tutorial series where you will create a tool called BAMStats and publish it onto Dockstore.
-
 Getting Started with Docker
 ===========================
 
 Tutorial Goals
 --------------
+This is the first part of our "Getting Started" tutorial series, where we will walk you through the creation of a tool called BAMstats and publish it onto Dockstore. This particular page, "Getting Started with Docker," will go through the basics of Docker and have you develop a Dockerfile for the BAMstats tool.
+
+This tutorial assumes you have basic knowledge of a Unix-like file system, such as working directories, moving between directories on the command line.
```
svonworl, line 8: Add "and" between "directories," and "moving"

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -1,36 +1,42 @@
...
--  Test Docker image locally
+-  Test your Docker image locally
 
 Introduction to Docker
 ----------------------
 
-.. note:: See `Docker Overview <https://docs.docker.com/get-started/overview/>`__ for an excellent overview about Docker.
+.. note:: See `Docker Overview <https://docs.docker.com/get-started/overview/>`__ for an excellent overview of Docker.
+
+There is a good chance that you have heard of containerization. A container is essentially an emulated computer system that bundles up kind of software. It's similar to a virtual machine, but is usually much lighter on system resources, as a container, unlike a VM, does not need to run an extra operating system.
```
svonworl, line 21: Reads rough around "up kind".

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -1,36 +1,42 @@
...
 
-Docker is a fantastic tool for creating light-weight containers to run
-your tools. It gives you a fast, VM-like environment for Linux where you
-can automatically install dependencies, make configurations, and setup
-your tool exactly the way you want, just as you would on a "normal"
-Linux host. You can then quickly and easily share these Docker images
-with the world using registries like Quay.io (indexed by Dockstore),
-Docker Hub, and GitLab.
+Docker images are usually shared on registries such as Quay.io, Docker Hub, and GitLab. When a user downloads an image and runs it locally, it creates an editable copy of the image. The editable copy is called a Docker container. You can think of an image as a template, and a container as something made from that template.
```
svonworl, line 25: You might wanna run this by Charles or someone else with more Docker experience than me.  I think of the Docker "image" as the bundled-up set of files, and the "container" as the running application spawned from the image within Docker.

aofarrel: I did a lot of research into this for the Dockstore dictionary. Here's my current understanding of the guts of Docker... Charles et al, let me know if this is technically correct, because I base multiple explanations here and elsewhere on this:

A Docker image is a read-only executable file. It is essentially a bundled-up filesystem that usually contains a piece of software and requirements of said software. In bioinformatics our Docker images usually use a Docker image of a Linux filesystem + some Linux distro's goodies (such as apt-get) as the base image, and then add layers on top of it. Every RUN, COPY, and ADD instruction in a Dockerfile adds a new layer on top of the base image/previous layers. That being said, the contents of a Docker image do not necessarily package anything **_from_** any particular operating system, but it does need things **_for_** a given operating system. This can be seen with the absolute minimal Docker image possible, [hello-world](https://registry.hub.docker.com/_/hello-world/?tab=tags), which consists of naught but scratch (the absolute most basic of base images) and [a single binary program](https://github.com/docker-library/hello-world/blob/master/amd64/hello-world/hello). In order to run that binary program correctly, there are different versions of the Docker image for different hardware architectures, as well as different host OSes, but even though they can fluctuate greatly in size for different host OSes/archs (the [smallest](https://registry.hub.docker.com/layers/hello-world/library/hello-world/latest/images/sha256-995efde2e81b21d1ea7066aa77a59298a62a9e9fbb4b77f36c189774ec9b1089?context=explore) is less than 3 KB, the [largest](https://registry.hub.docker.com/layers/hello-world/library/hello-world/latest/images/sha256-1e0054d86f2721056c9db5c2322fb8883d185fa1165bca69cb0ac88eeb40be30?context=explore) is over 100 MB) none of them contain much of anything from any particular operating system; they just need a different setup depending on what host OS/arch they are going to be running on, in the same way that [Debian for arm64](https://cdimage.debian.org/debian-cd/current/arm64/iso-cd/) isn't the same size as [Debian for i386](https://cdimage.debian.org/debian-cd/current/i386/iso-cd/).

When it's time to actually run the software package inside a Docker image, the Docker engine creates an editable copy of the Docker image. This editable copy is a Docker container. (My object-oriented Python brain compares this to an object being an instance of a particular class' non-editable-at-runtime template.)

A Docker image:
* cannot be edited, but a new version can be created via a Dockerfile
* typically bundles up a filesystem, but not an entire operating system, yet might bundle components of an operating system
* is created using `docker build` and a Dockerfile via a process that involves building layers on top of each other

A Docker container:
* can be edited in the sense that you can add/delete files from its filesystem while it is being executed
* mounts its own filesystem separate from the host OS, usually based upon the filesystem provided in the Docker image (the user can mount additional stuff into this filesystem from the host OS using the volume/mount commands)
* runs its own processes in such a way that it can only see said processes, not other stuff going on in the host machine, and with unique user ID(s)
* shares the same kernel as the host OS
* does not have a guest OS; what is being run in a Docker container is technically being run "natively" on the host OS

A Docker tag:
* is a mutable reference to a particular Docker image

A Docker digest:
* is an immutable description of a particular Docker image derived from the `docker build` command that created said 
Docker image
* can be used to reference a particular Docker image more precisely than a tag

The Docker engine:
* Is not a hypervisor
* However, it can manage several Docker containers at the same time, like how a hypervisor can manage several guest OSes at the same time

denis-yuen: I'm more or less ok with most of this except that "editable copy" is kind of an unusual way of putting it. While true, it seems incomplete? https://docs.docker.com/get-started/overview/#containers has "runnable instance"  which is more familiar to me

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -1,36 +1,42 @@
...
+
+Users of HPC (high performance compute) systems may not be able to run Docker Engine directly, depending on your sysadmin's policies and the details of your HPC's scheduling system. We can't go over all of possible HPC setups here, but we recommend checking out `Singularity <https://sylabs.io/guides/2.6/user-guide/singularity_and_docker.html>`__ and `Shifter <https://github.com/NERSC/shifter>`__, both of which are designed for running Docker images without using the Docker Engine program. If your HPC does not support Docker Engine, there is a strong chance it is already set up with an alternative.
```
coverbeck, line 31: > high performance compute

High Performance Compute or High Performance Computing

docs/getting-started/getting-started-with-docker.rst
aofarrel: Given the fact that a Docker image is read-only, I like mentioning that it is editable, but honestly my OOP brain just thinks of it as an instance. Think I'll reword the paragraph this shows up in slightly to get those points across.

docs/getting-started/getting-started-with-cwl.rst
aofarrel: Looks like I used second-person implicit for the first part, then came back to the sentence and tacked on the second part without remembering the object and object-verb of the first part was implicit... hmmm

> This tutorial assumes [you have] basic familiarity... and have access..."

Curious if what I wrote is "grammatically correct but bad" vs "definitely incorrect," but either way, it's super confusing. Fixed it, thanks for catching this.

docs/getting-started/getting-started-with-docker.rst
```diff
@@ -22,13 +22,13 @@ There is a good chance that you have heard of containerization. A container is e
 
 Docker is a well-known type of containerization software. It provides a fast environment for both users and developers. A developer can package software and its dependencies into an Docker image, and then share that Docker image with users. Users who download your Docker image and run it with the Docker program will be able to run the software you packaged without having to handle the installation of your program's prerequisites or anything else; it's ready to go as-is.
 
-Docker images are usually shared on registries such as Quay.io, Docker Hub, and GitLab. When a user downloads an image and runs it locally, it creates an editable copy of the image. The editable copy is called a Docker container. You can think of an image as a template, and a container as something made from that template.
+Docker images are usually shared on registries such as Quay.io, Docker Hub, and GitLab. When a user downloads an image and runs it locally, the Docker Engine spins up an instance of the Docker image that can be edited and has its own filesystem. This instance is called a Docker container. You can think of an image as a template, and a container as something made from that template.
```
coverbeck, line 25: I know this has been discussed a bit, but I still have a hard time think of a container being an image that you edit. What if you just remove `that can be edited and has its own filesystem`, and punt on the rest?

coverbeck (approved):

PR #211, opened 2022-07-07
Initial Dockstore Logo Usage Guidelines commit

bvizzier:
Additions:
- Dockstore Logo Usage Guidelines initial commit.

Changes: 
- I was receiving an error message about "language = None" in the conf.py file.  Per the message, I changed it to "language = 'en'".
- The makefile change was due to an error I was getting on macOS.

I would like to find a higher contrast background color for the white logo variants in the table but I do not think it is a blocker for this document.  Thoughts?


bvizzier:
Can be viewed at:  https://docs.dockstore.org/en/dock-2180_logo_use/dockstore-logo-usage-guidelines.html

bvizzier:
Just pushed a change to include this page in the Discourse exclude list.

docs/Makefile
```diff
@@ -13,7 +13,7 @@ help:
 
 glossary:
 	echo "\033[1mGenerating dictionary.rst...\033[0m"
-	python _attic/glossary_generator.py
```
denis-yuen, line 16: FYI @aofarrel 
Was this causing an error @bvizzier ?

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+* Use the logo or icon color that works best in the context of the document, web page, or application.
+* The Dockstore logo is preferred.  Use the Dockstore icon only when space is limited or when it is clear that the reference is to Dockstore.
+* Spacing:
+  * In most cases, the logo should be of equal prominence to any other brands with which it is shown.
+  * Allow for adequate space around the logo when used in conjunction with other identifiers.
+  * Do not place other identifiers closer than the equivalent of the width of 1/2 the logo.
+
+Do Not:
+
+* Don’t place the logo on clashing colored backgrounds.
```
denis-yuen, line 124: if sticking with the periods, remove the "don't" at the start of the lines
Here is an example of bullet lists that need a period to complete the introductory stem sentence:

docs/Makefile
bvizzier: Yes.  It was erroring out on the python invocation.  Changing it to python3 allowed it to work.

aofarrel: My hunch is that Ben's `python` points to python2, while mine points to python3. I changed mine to 3 a while back and forgot about it, so this was a good catch on Ben's part -- calling Python3 explicitly is the way to go.

As a side note, Mac OS is going to finally remove Python 2 in the near-ish future: https://forums.macrumors.com/threads/apple-finally-removing-python-2-in-macos-12-3.2332855/

docs/dockstore-logo-usage-guidelines.rst
aofarrel: I have the opposite interpretation of the same link. Yeah, it talks about using bullets to close off introductory stem sentences, but that's not the only situation that it cites.

> * Use a period or other full stop after every bullet that is a sentence. (Just like this bullet list you’re reading now.)
> * Use a period after bullet list that completes the opening stem sentence that introduces it.
> * Don’t use a period after bullet lists that are not complete sentences or do not complete the opening stem sentence. 

These are full sentences. The subject is implicit, but it's still a subject.

docs/Makefile
denis-yuen: ok, good to know. 
I guess in the end, CI wins anyway

bvizzier: macOS has already removed Python 2 and the /usr/bin/python symbolic link in the 12.3 update.   The official command is now /usr/bin/python3.

docs/dockstore-logo-usage-guidelines.rst
aofarrel: Unless what you're saying is that the introduction of periods make them look as though they are ending off introductory stem sentences? Even that I'm not so sure about, since Do Not being capitalized doesn't follow proper sentence capitalization anyway.

docs/Makefile
aofarrel: Just checked .bashrc and I think I discovered why I didn't get the memo... oops!
alias

docs/dockstore-logo-usage-guidelines.rst
bvizzier: I have altered the wording in the Do & Do Not lists to make it more consistent.  I'll push it shortly.

bvizzier: BTW:  I have updated the bullet list to the following.  I'll push the change once we have the broken link resolved elsewhere.

Do:

- Use the logo or icon color that works best in the context of the document, web page, or application.
- Use the Dockstore logo when space permits. Use the Dockstore icon only when space is limited or when it is clear that the reference is to Dockstore.
- Provide the Dockstore logo equal prominence to any other brands with which it is shown.
- Allow for adequate space around the logo when used in conjunction with other identifiers.

Do not:

- Place the logo on clashing colored backgrounds.
- Place the reverse logo on similar tints.
- Place any logo variation on busy backgrounds.
- Place special effects such as drop shadows on the logo.
- Skew or distort the logo.
- Alternate the color of the logo.
- Place other identifiers closer than the equivalent of the width of 1/2 the logo.

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+  :width: 64
+
+.. |Icon White| image:: /assets/images/Dockstore_Logos_and_Icons/Dockstore-icon-white.png
+  :width: 64
+
+===============================
+Dockstore Logo Usage Guidelines
+===============================
+
+About
```
aofarrel, line 49: I'd replace this heading with "Background"

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+=====
+It is important to establish a sense of branding throughout our publicly facing, digital media. Currently, we have several themes / layouts, navigation systems, font sets, color palettes. We would like to unify our designs under a single set of standards.
```
aofarrel, line 51: I'd suggest:
> Currently, we have several themes, layouts, navigation systems, font sets, and color palettes.

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+      - Dockstore-logo-vertical-black-blue
+      -
+    * - |Horizontal white blue|
+      - |Vertical white blue|
+      -
+    * - Dockstore-logo-horizontal-white-blue
+      - Dockstore-logo-vertical-white-blue
+      -
+
+A zip file containing these logos in both PNG and SVG formats can be found :download:`here </assets/images/Dockstore_Logos_and_Icons.zip>`.
```
aofarrel, line 109: If we're going to make high-res logos, they probably belong here, and a note should mention it includes high-res.

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Use
+===
+Do:
+
+* Use the logo or icon color that works best in the context of the document, web page, or application.
+* The Dockstore logo is preferred.  Use the Dockstore icon only when space is limited or when it is clear that the reference is to Dockstore.
+* Spacing:
```
aofarrel, line 117: I know you mentioned you updated the formatted list from what's shown here, but in case you want to switch back to nested lists, you can try adding an extra newline after each bullet (including the parent)

Screen Shot 2022-07-07 at 4 13 34 PM

docs/dockstore-logo-usage-guidelines.rst
aofarrel: Some comments on these Do Nots...

> the reverse logo on similar tints

I'd recommend saying white-on-transparent instead of reverse logo just for clarity's sake. And I'd argue this applies to more than just the reverse logo. It might be best to reword this as "Place the logo on backgrounds where it is hard to read"

> Alternate the color of the logo.

I'd recommend "Alter" instead of "Alternate"

> Place other identifiers closer than the equivalent of the width of 1/2 the logo.

I like this in principle, but we're being a little hypocritical here.
Screen Shot 2022-07-07 at 4 26 12 PM

aofarrel (changes requested):
Mostly nitpicks, but I do have some feedback on the images themselves. IMHO, 253 × 147 is just not big enough in this day and age. If this is the go-to place for the Dockstore logo, it'd be best to include one that looks nice on a Google Slide.

Suggestions:
* Upscale our logos to at least 800x600-ish, and have the upscaled logos on this page
* Upscale our logos, but keep the small ones in the table and only put the HD ones in the download
* Convert our logos to SVG so we can scale them easily (con: SVGs are not as widely supported as PNGs)

docs/dockstore-logo-usage-guidelines.rst
bvizzier: The SVG files are in the zip file along with the PNG files.

Why 800x600?   That seems large to me for most use.  I can scale them up, but, then what is the magic size.

The horizontal ones are 406 x 97.
The vertical ones are 253 x 147.
The icons are 140 x 97.

I do want to note that the images are scaled down in the table to get them to fit comfortably in the table format within the fixed width window.

docs/dockstore-logo-usage-guidelines.rst
bvizzier: This was a mistake.  Accidentally deleted the blank line.

docs/dockstore-logo-usage-guidelines.rst
bvizzier: Changed paragraph.

coverbeck:
If you want this to be released sooner that 1.13, the PR should target `hotfix/1.12.2` instead of `develop`. Not saying you should necessarily, but I wanted the options to be clear. There is no reason to wait until 1.13, OTOH, is there a huge demand for this? Your call.

docs/Makefile
coverbeck: Interesting, I use a Python 3 virtual environment, and it works with `python`. But going to defer to the more experienced Python developers.

docs/dockstore-logo-usage-guidelines.rst
coverbeck: This whole section is presumably for external, non-Dockstore sites, but this is talking about "our" media, our themes, and how we aspire to unify them. While it's true that we want to unify them, I don't understand why we're talking about that here.

I would instead just say something like, very roughly/poorly, needs a lot of rework: Do you want to acknowledge or credit Dockstore on your site? Here are some guidelines for..., etc.

Or maybe I'm misunderstanding the intent of this ticket.

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+  * Do not place other identifiers closer than the equivalent of the width of 1/2 the logo.
+
+Do Not:
+
+* Don’t place the logo on clashing colored backgrounds.
+* Don’t place the reverse logo on similar tints.
+* Don’t place any logo variation on busy backgrounds.
+* Don’t place special effects such as drop shadows on the logo.
+* Don’t skew or distort the logo.
+* Don’t alternate the color of the logo.
```
coverbeck, line 129: Does this mean if you have two logos, they should have the same color? Or does it mean not change the color of the logo at all?

docs/dockstore-logo-usage-guidelines.rst
```diff
@@ -0,0 +1,129 @@
...
+* Spacing:
+  * In most cases, the logo should be of equal prominence to any other brands with which it is shown.
+  * Allow for adequate space around the logo when used in conjunction with other identifiers.
+  * Do not place other identifiers closer than the equivalent of the width of 1/2 the logo.
+
+Do Not:
+
+* Don’t place the logo on clashing colored backgrounds.
+* Don’t place the reverse logo on similar tints.
+* Don’t place any logo variation on busy backgrounds.
```
coverbeck, line 126: Not sure I get this one. Does it imply you can place logo variation on quiet backgrounds? And what does "place logo variation" mean?

bvizzier: We have 13 variations:  horizontal blue, vertical blue, blue icon, horizontal black, vertical black, etc.

I could replace "place" with "use."

If we want to require the use of blue on white we will need to update our products to comply with that.

docs/dockstore-logo-usage-guidelines.rst
bvizzier: Should read "Don't alter the color of the logo."  Thanks to Ash, this was changed in a subsequent push.

bvizzier: The first paragraph has been reworked.

denis-yuen:
> * Convert our logos to SVG so we can scale them easily (con: SVGs are not as widely supported as PNGs)

we have SVG and EPS here
 https://github.com/dockstore/dockstore-ui2/tree/develop/images/high-res

docs/dockstore-logo-usage-guidelines.rst
bvizzier: Yes, we are in violation of the UCSC logo usage guidelines definitely in 1 area and possibly in a second:  1) [wrong logo](https://communications.ucsc.edu/visual-design/logos-imagery/), 2) spacing requirements. 

There is an interesting conflict here and that is between the spacing requirements and the equal prominence requirement.  I would say this usage meets the equal prominence requirement.

docs/dockstore-logo-usage-guidelines.rst
bvizzier: If we are going high-res we should probably also do small, at least for the icon.  Small tends to be more problematic than large due to the loss of visual elements when downscaling.

bvizzier:
The Dockstore logo in the dockstore-ui2 folder is the old icon (with the tabs) and is only in PNG.  

As noted, the SVG files are in the zip.  I'm moving the link to the zip file to above the table instead of below it because that is being missed.

docs/dockstore-logo-usage-guidelines.rst
denis-yuen: Well, they are guidelines :wink:
5af1800dbd18b30ced7a98fd63f0fe1f411091a0da880f6bf4dcdd967cce8270

bvizzier: I've removed the "Place other identifiers closer than the equivalent of the width of 1/2 the logo" guideline.

denis-yuen (approved):

denis-yuen (changes requested):
guess we have to wait for builds to pass

bvizzier:
> 

I can cherry-pick it once it is approved.  It's not like it is a big change.

aofarrel (approved):
That zip is quite plentiful now. Appreciate it!

PR #213, opened 2022-07-18
Workaround for merge 1.12.2 and develop merge conflicts

aofarrel:
Pulling in this PR *should* solve #212's merge conflicts, which I currently cannot do as 1.12.2 is a protected branch that I cannot commit to directly.


aofarrel:
Charles handled this on his end as he has the permissions necessary to do this using just hubflow.

PR #216, opened 2022-07-20
Add links to return to the top of long documentation pages

aofarrel:
This is something I've been thinking of adding for a while (see #191). Seeking feedback on whether ▲  or a more explicit `(back to top)` would be preferred. This PR currently has both options for comparison.


aofarrel:
Side note: Seeking some docs on adding discourse topics. This page would likely benefit from one.

denis-yuen:
> Side note: Seeking some docs on adding discourse topics. This page would likely benefit from one. I know what it is, just unsure if I need to do something besides adding a flag like discourse-id NNNN at the bottom of the page.

We have a github action now https://github.com/dockstore/dockstore-documentation/actions/workflows/add-discourse-topic.yml which proposes a topic for new pages. I think its fairly safe to run on PRs. 

That said, it looks like this page already has one

kathy-t:
> Seeking feedback on whether ▲ or a more explicit (back to top) would be preferred. This PR currently has both options for comparison.

Looks like the PR still has both options, should decide on one so it's consistent

denis-yuen:
Screenshot from 2023-04-25 10-18-00

Comparing the two, the arrow seems easy to overlook/misunderstand.

kathy-t (approved):

fhembroff:
Yeah, comparing the two, I prefer the "(back to top)".

fhembroff (approved):
Looks good!

PR #221, opened 2022-09-10
Move dockstore.yml templates up in our sidebar

aofarrel:
I got lost on a documentation website whose index I rewrote. I have only my hubris to blame. Now, it's time to make things right.


denis-yuen (approved):

coverbeck (approved):
Beyond the scope of this PR, but just noticed it and the title "Templates", which appears in the sidebar is pretty vague, IMHO.

PR #222, opened 2022-09-10
Fix glossarpy sometimes writing an absolute path

aofarrel:


aofarrel:
I did regenerate the files but forgot to commit them — will do that tomorrow/soon.

denis-yuen (approved):
We probably need to do something about the branches in this repo in regards to 1.13 but I think we have tickets for that already

aofarrel:
Another possibility: https://github.com/dockstore/dockstore/issues/5128#issuecomment-1242782166

Let me know if you'd prefer that in this PR.

coverbeck (approved):
> We probably need to do something about the branches in this repo in regards to 1.13 but I think we have tickets for that already

I was unaware there was a hotfix branch as well. We should do a `hf finish` after this merged.

PR #224, opened 2022-09-13
DOCK-2207: Clarify requirements for .dockstore.yml

svonworl:
Modifies the documentation to:
1. Note restrictions on characters in workflow/tool names.
2. Specify absolute paths for descriptor and test file paths. 
3. Add "absolute path", "primary descriptor file", and "secondary descriptor file" entries to the Dictionary.

I consulted with Ash, and we agreed that, perhaps, there wasn't enough confusion about the workflow file extension to merit documentation enhancements.

There are some rumblings in the ticket comments about whether to specify that a test parameter file is a hard requirement.  IMHO, it's sort of a gray area - we ask for a test parameter file, but it can be functionally empty - so I left the docs as-is with respect to that...

https://ucsc-cgl.atlassian.net/browse/DOCK-2207


docs/_attic/glossary_entries.py
```diff
@@ -22,6 +22,12 @@
 AGC = GlossEntry("AGC",
 	acronym_full="[Amazon Genomics CLI]")
 
+AbsolutePath = GlossEntry("absolute path",
+	definition="A path that starts with the character ``/`` and contains the full set of directories necessary to resolve a file, starting from the root directory. For example: ``/Dockstore.cwl`` or ``/bin/sh``",
```
coverbeck, line 26: > from the root directory

I'm not sure how to work it in, and especially here where it's a generic definition, but the absolute path in the context of the bug is "relative" :) to the repository's root, not the (local) file system root.

coverbeck (approved):

coverbeck (approved):

docs/_attic/glossary_entries.py
```diff
@@ -475,6 +481,20 @@
 	pronunciation='', 
 	seealso="[spot instance]")
 
+PrimaryDescriptorFile = GlossEntry("primary descriptor file",
+	acronym_full="",
+	definition="The main [descriptor file] that describes a workflow or tool.",
```
denis-yuen, line 486: main and primary seem like synonyms that aren't really shedding much light here

denis-yuen: maybe something like "the initial descriptor file that Dockstore starts from to find other descriptors"

coverbeck:
@svonworl FYI, retargeted the PR to release/1.13.0.

denis-yuen (approved):

aofarrel (commented):
Couple notes...
* The naming workflows page will not show up on the sidebar nor the homepage. If that's intended, no worries.
* Consider committing the generated files, because if you don't, someone else probably will do so on another branch and then things will get confusing.

svonworl:
> Couple notes...
> 
> * The naming workflows page will not show up on the sidebar nor the homepage. If that's intended, no worries.
> * Consider committing the generated files, because if you don't, someone else probably will do so on another branch and then things will get confusing.

* "Naming workflows" has the same visibility as before, which is intended, I assume.
* Thanks for the heads up regarding the generated files.

aofarrel (approved):
Looks great, thanks!

PR #226, opened 2022-09-14
DOCK-2197: Add Galaxy information

svonworl:
Added miscellaneous Galaxy-related information:

Questions/notes:
1. Added a subsection related to test parameter files to "Getting Started with Galaxy".  It references some instructions on usegalaxy.org which tell the user how to create a parameter file with a format that seems to match the test parameter files of existing Dockstore-registered Galaxy workflows.  So, this is probably the correct advice, but I am far from a Galaxy expert, so if you know for sure, please confirm.  I believe usegalaxy.org is one of the official Galaxy sites, but is there a more canonical url?  I could not find one.
2. There is already a link to "Launch With Galaxy" at the top of the "Getting Started with Galaxy" page, so I did not add another (as requested by the ticket).
3. The Terra "create a galaxy env" interface changed significantly, so I updated the instructions, in addition to the screenshot.

https://ucsc-cgl.atlassian.net/browse/DOCK-2197


denis-yuen:
Is actually https://github.com/dockstore/dockstore/issues/5006 (makes it easier to match PRs with issues)

denis-yuen:
> 1. usegalaxy.org is one of the official Galaxy sites, but is there a more canonical url?

Yeah, I think this is their main US site. https://help.galaxyproject.org/t/difference-between-usegalaxy-org-and-usegalaxy-eu/4722/4

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -72,6 +72,11 @@ Upload the workflow to GitHub
 
 These steps are outlined `here. <https://docs.github.com/en/github/managing-files-in-a-repository/adding-a-file-to-a-repository>`__
 
+Add Test Parameter File
+-----------------------
+
+You can add a test :ref:`dict parameter file` to your source code repository, allowing other users to easily launch a test run of your workflow.  Follow `these instructions <https://usegalaxy.org/training-material/topics/contributing/tutorials/create-new-tutorial-technical/tutorial.html#testing-the-workflow-recommended>`__ to create a test parameter file.  Then, add the file you have created to your GitHub repository as you did with your '.ga' Galaxy workflow file above.
```
denis-yuen, line 78: Was like this already, but is it possible to steer users toward the newer .yaml Galaxy workflow language? https://github.com/galaxyproject/gxformat2

(more human readable, writeable, etc. )

denis-yuen (commented):
Cannot comment on it directly, docs/assets/images/docs/galaxy/create_galaxy_environment.png looks a lot smaller in terms of pixels and is 50% smaller in size. How does it look in context?

docs/_attic/glossary_entries.py
```diff
@@ -193,6 +193,13 @@
 	institute="", 
 	pronunciation="")
 
+DRS = GlossEntry("DRS",
+        acronym_full="Data Repository Service",
+        definition="A standardized [API], created by the [GA4GH] Cloud Work Stream, that provides portable access to a repository of data resources.",
```
coverbeck, line 198: I would say "to repositories of data resources".

And/or, from https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.0.0/docs/, "a generic interface to data repositories".

docs/_attic/glossary_entries.py
```diff
@@ -514,6 +521,13 @@
 	pronunciation='', 
 	seealso="[workflow]")
 
+TRS = GlossEntry("TRS",
+        acronym_full="Tool Registry Service",
+        definition="A standardized [API], created by the [GA4GH] Cloud Work Stream, that provides portable access to a registry of tools, workflows, and associated files.  Every resource in a TRS registry has a public ID that can be used to retrieve it.  Dockstore `provides <https://dockstore.org/api/static/swagger-ui/index.html#/GA4GH>`__ a TRS interface to its database.",
```
coverbeck, line 526: The link is pointing to the TRS beta implementation; it should be https://dockstore.org/api/static/swagger-ui/index.html#/GA4GHV20

Arguable nit: "...to its database", seems like an implementation detail, unless you're using database in a very general sense, but I don't have a better suggestion off-hand, so you don't have to change it if you like it.

docs/dictionary.rst
```diff
@@ -386,6 +388,22 @@ DOI
 
 
 
+.. _dict DRS:
+
+DRS
```
coverbeck, line 393: Beyond the scope of this PR, but I'd suggested before and wonder again if this and other generated files shouldn't be in .gitignore. I started to review this before realizing I already had reviewed what this is generated from.

coverbeck (approved):
Good move to check the current implementation on Terra and update accordingly.

svonworl:
> Cannot comment on it directly, docs/assets/images/docs/galaxy/create_galaxy_environment.png looks a lot smaller in terms of pixels and is 50% smaller in size. How does it look in context?

Looks fine in my environment. The new file is about half the size in bytes, at the expense of maybe a little less detail for Retina viewers.

docs/getting-started/getting-started-with-galaxy.rst
svonworl: Done, I basically added a "Note" at the end of the section that explains how to export, explaining "Format 2", and that if the interface allows it, you should export your workflow as a "Format 2" file.

docs/_attic/glossary_entries.py
svonworl: Link fixed.  Re "to its database", yes, I was using that generically, rather than as a reference to PostgresQL or whatever.  It didn't add much, so I deleted it, thus avoiding the need to consider whether it's common usage, should be replaced, etc.

denis-yuen (approved):
looking good, may have build issue to deal with

PR #230, opened 2022-11-22
Update information on Dockstore CLI

aofarrel:
We don't quite have a singular piece of documentation that I can easily point to when I say "wanna run this cool WDL I wrote locally ~because you are more likely to get away with your laptop slowly pinging SRA 10000 times one-at-a-time rather than Terra pinging SRA 10000 times all-at-once~? here's what the Dockstore CLI actually is and how to install it." 

This PR... does not address that, but it makes the ground fertile for such a change later.

Changes included:
* added cwltool, miniwdl, and environment variable to dictionary
* slightly less heavy focus on cwltool in our Dockstore CLI docs
* rework our FAQ entry on Dockstore CLI for Mac entirely
  * now two entries: one general, one for the permissions error
  * explicitly notes that our docs have Ubuntu in mind, not Mac
  * warn people that Catalina and onward uses zsh
  * provide two possible solutions for the Mac Docker permissions error
  * link to dictionary entry for environment variables
* small updates to the meta documentation of the dockstore dictionary
* improve an example in our docs by actually linking to the workflow being talked about

No NCBI servers were harmed in the making of this pull request. Probably.


aofarrel:
CI error is about discourse. I think the page in question deserves a discourse topic page, but I don't know how to actually add one... please amend this PR to remedy that if possible.

docs/_attic/glossary_entries.py
```diff
@@ -241,6 +254,13 @@
 	institute="Dockstore", 
 	pronunciation='')
 
+EnvironmentVariable = GlossEntry("environment variable", 
+	acronym_full="", 
+	definition="A variable that affects how processes run on a computer. For example, [cwltool] references the environment variable $TMPDIR when deciding where to place files. You can set environment variables in Bash using the `export` command. Environment variables are sometimes called env var or env for short.",
```
coverbeck, line 259: "...are sometimes called env var..." -- I've never heard that, but it might just be me.

docs/_attic/glossary_entries.py
```diff
@@ -241,6 +254,13 @@
 	institute="Dockstore", 
 	pronunciation='')
 
+EnvironmentVariable = GlossEntry("environment variable",
```
coverbeck, line 257: I wonder if this is getting a little too low-level. We have to assume some knowledge, and, for example, I think the sentence you have below, "You can set environment variables in Bash using the export command" while not wrong, also isn't precise:

* It's not a Bash-only feature
* Using export isn't the only way to set them.

But more broadly, I don't think we should be discussing different shells, the difference between SET and EXPORT, etc. in the Dockstore documentation.

Just my opinion, feel free to push back.

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -6,10 +6,10 @@ For general FAQs not related to the Dockstore CLI, please see :doc:`our main FAQ
 .. contents:: Table of Contents
   :local:
 
-How does launching CWLs with Dockstore CLI compare with cwltool?
-----------------------------------------------------------------
+How does launching CWLs/WDLs with Dockstore CLI compare with cwltool/Cromwell?
+------------------------------------------------------------------------------
 
-Under the hood, the Dockstore CLI invokes cwltool to launch CWL workflows. However, it adds additional features. The Dockstore CLI can generate JSON parameter files from
+Under the hood, the Dockstore CLI invokes cwltool to launch CWL workflows. Likewise, the Dockstore CLI invokes Cromwell to launch WDL workflows. However, it adds additional features. The Dockstore CLI can generate JSON parameter files from
```
coverbeck, line 12: The Dockstore CLI invokes cwltool by default, but it is configurable: https://docs.dockstore.org/en/stable/advanced-topics/dockstore-cli/advanced-features.html?highlight=cwl%20runner#alternative-cwl-launchers

There is no such option for WDL, although if/when we add miniwdl support, that will probably change.

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -22,29 +22,50 @@ for more information on these two file transfer sources.
...
+
+Why am I getting "Mount denied" errors when launching workflows on Mac?
+-----------------------------------------------------------------------
+Docker behaves a bit differently on a `Mac <https://docs.docker.com/docker-for-mac/osxfs/#/namespaces>`__ than on a typical Ubuntu machine. On a Mac, by default, the only shared volumes Docker can access are /Users, /Volumes, /tmp, and /private. This is problematic because many workflow executors, such as `cwltool`, uses your TMPDIR (the :ref:`dict environment variable`) to setup volumes with Docker. Since `$TMDIR` on a Mac can default to a subdirectory of /var, which Docker usually cannot access, you may run into an error. When using cwltool, the error will look something like this:
```
coverbeck, line 39: "to setup volumes with Docker" -> "to set up volumes with Docker" -- I had to look it up, but setup is one word when used as a noun, but two words when "set" is a verb.

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -22,29 +22,50 @@ for more information on these two file transfer sources.
...
+You can configure shared paths from Docker -> Preferences... -> File Sharing.
+See https://docs.docker.com/do.
+ERRO[0000] error getting events from daemon: net/http: request canceled 
+```
+
+This issue can happen when using cwltool or Cromwell, with or without the Dockstore CLI. There are two possible ways to fix this. 
+
+Option A: Overwrite $TMPDIR
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can set the $TMPDIR :ref:`dict environment variable` to be under one of the shared volumes that Docker for Mac can access by default. You can do this by doing something similar to the following:
```
coverbeck, line 55: I've never done this, but I assume that you need to do this before you start Docker? If so, can you add the clarification?

docs/dictionary.rst
```diff
@@ -25,6 +25,7 @@ Dockstore Dictionary
 	* :ref:`dict container`
 	* :ref:`dict Cromwell`
 	* :ref:`dict CWL`
+	* :ref:`dict cwltool`
```
coverbeck, line 28: I wish we could .gitignore this file, and maybe we can. We'd have to check how the builds work and whatnot.

No action required, just noting. But I started to review this file before realizing I'd already reviewed the files it's generated from. It's a generated file, so it shouldn't show up in reviews.

coverbeck (commented):
> CI error is about discourse. I think the page in question deserves a discourse topic page, but I don't know how to actually add one... please amend this PR to remedy that if possible.

I can do that, but not until after Thanksgiving :(

docs/_attic/glossary_entries.py
denis-yuen: (also no) 
https://books.google.com/ngrams/graph?content=%22env+var%22%2C%22environment+variables%22&year_start=1800&year_end=2019&corpus=26&smoothing=3

aofarrel: I've seen "env variables" and "environment vars" and "envs" -- wanted to get all of those across. But yeah, now that I think about it I've never seen "env vars." Is the shorthand self-evident enough for me to get rid of that bit or should I list the variations that do get use?

docs/_attic/glossary_entries.py
aofarrel: Almost all of our documentation assumes a bash-on-Ubuntu environment (it's just that bash-on-Mac and zsh-on-Mac are 99% the same), so I think using that as an example of an env variable is the best example. Good point about set and export though, I'll see what I can think of...

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
aofarrel: Good catch!

docs/dictionary.rst
aofarrel: Think I found a solution: https://github.com/dockstore/dockstore-documentation/pull/231

docs/_attic/glossary_entries.py
denis-yuen: I think environment variables is probably enough. If they've heard of one of the more unusual variations, hopefully they already learned what it is.

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -22,29 +22,50 @@ for more information on these two file transfer sources.
...
-    env variable) to setup volumes with Docker, which on a Mac can default
-    to a subdirectory of /var. In order to get ``cwltool`` working on your
-    Mac, you need to set your TMPDIR to be under one of the shared volumes
-    in Docker for Mac. You can do this by doing something similar to the
-    following:
-    ::
+.. note:: Our tutorials all assume you are using bash on an Ubuntu machine. Beginning with Mac OS 10.15 Catalina, the default shell has changed from bash to zsh. In almost all cases (especially in our tutorials), syntax between bash-on-Ubuntu, bash-on-Mac, and zsh-on-Mac is the same, but it is possible for slight differences to arise.
 
-        export TMPDIR=/tmp/docker_tmp
+See `Docker for Mac <https://docs.docker.com/engine/installation/mac/>`__ for information on installing Docker for Mac. With the exception of some installation and possible folder permission differences (see next FAQ below), usage of the Dockstore CLI on a Mac is more or less the same as it is on Ubuntu.
```
aofarrel, line 27: This links to Docker Engine while our quick start guide currently links to a page that will suggest Docker Desktop.

denis-yuen: We may not have a choice for Mac. While Ubuntu has both choices (Docker Engine unhelpfully called "Docker server" here https://docs.docker.com/engine/install/#server ) I'm operating under the assumption Mac users don't have the choice.

denis-yuen (approved):
Oops, missed that I didn't approve. 
Approved with the comments above

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
aofarrel: I don't know for sure but I will add that in.

coverbeck (commented):
Looks great! But we still need a Discourse topic... I realized now I'd said I'd add one, but totally forgot. However, since you first created this PR, @svonworl has added topics to all pages missing them. Can you please rebase/merge from develop to this branch first and see if the tests pass?

If they don't, there is a way to automate the adding of a discourse topic that we can get into next.

denis-yuen:
Looks like builds pass now, tagging @aofarrel since you can't do a proper review as an author
@svonworl and @coverbeck for Mac-stuff among other stuff

svonworl (approved):

hyunnaye (approved):

coverbeck (approved):

PR #247, opened 2023-05-26
Feature/back to top links

aofarrel:
Builds off of https://github.com/dockstore/dockstore/issues/4830, addresses https://github.com/dockstore/dockstore/issues/5510. 

Changes:
* Adds back-to-top links to our other FAQs
* Fixes some missing back-to-top links on the GHA FAQ, and makes their spacing more consistent
* Fix a minor typo

Notes: 
* I'm targeting 1.14 since I think we can squeeze it in, but feel free to change the destination branch if needed.
* There is a temporary issue with my local version of git preventing me from pushing to large repos, so commits were made directly on GitHub (but only after verifying everything compiled as expected locally).

### DIFF
--- a/docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
+++ b/docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
@@ -1,3 +1,5 @@
+.. _topCLIFAQ:
+
 Dockstore CLI FAQ
 =================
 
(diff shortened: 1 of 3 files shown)


aofarrel:
Currently making some additional changes to the FAQs but they will be in a different PR.

denis-yuen:
> * There is a temporary issue with my local version of git preventing me from pushing to large repos, so commits were made directly on GitHub (but only after verifying everything compiled as expected locally).

Huh, that's an interesting one. Can chat on  #dockstore-random

denis-yuen:
Wondering why so many extra commits show up, however, the diff looks ok

coverbeck (approved):

PR #248, opened 2023-05-26
Assorted FAQ improvements

aofarrel:
Builds upon #247. This is bigger than #247, so it might be best to aim this one at 1.15?

* Move environment question from main FAQ to advanced CLI FAQ
  * A new user not very familiar with Dockstore might go the main FAQ, see in the first entry that we use 96 gigs of RAM to test tools, incorrectly conclude that's what they need to do anything with tools, and write off Dockstore as something for people with very beefy machines
* How to handle big reference files gets its own answer now, and that answer has changed because it turns out localizing [a 14 GB reference genome](https://console.cloud.google.com/storage/browser/_details/topmed_workflow_testing/tb/ref/index_decontamination_ref_output/Ref.remove_contam.tar;tab=live_object) into [thousands of Docker images](https://github.com/aofarrel/SRANWRP/blob/main/inputs/tb_accessions/tb_a3/tb_a3.txt) via [a surprisingly slow gsutil command](https://support.terra.bio/hc/en-us/community/posts/10732090690843-Switch-Localization-Delocalization-From-gsutil-to-gcloud-storage) is kind of awful
* Pull back on the CWL focus
* Add glossary entry for parent image
* Tweak sizes of screenshot images in the how-to-view-logs question
* Warn against using alpine as a parent image
  * Cromwell (and miniwdl, I think?) forces /bin/bash as an entrypoint so they don't work with WDLs
* Modify an answer that implied you cannot grab outputs that aren't in the workdir
* Small wording/punctuation changes


docs/_attic/glossary_entries.py
```diff
@@ -506,6 +506,10 @@
 	institute="", 
 	pronunciation='')
 
+ParentImage = GlossEntry("parent image",
```
denis-yuen, line 509: Can use ubuntu:22.04 , 18 was EOL in April

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -103,7 +113,23 @@ If a Docker lockup happens, you will notice in-progress WDL tasks do not progres
...
 
-To prevent these issues from happening, we recommend setting up your Cromwell configuration file to limit how many scattered tasks run at once, and then setting up the Dockstore CLI to make use of that Cromwell configuration file. :doc:`A step-by-step tutorial is available here. </advanced-topics/dockstore-cli/local-cromwell-config>`
+To prevent these issues from happening, we recommend setting up your Cromwell configuration file to limit how many scattered tasks run at once, and then setting up the Dockstore CLI to make use of that Cromwell configuration file. :doc:`A step-by-step tutorial is available here. </advanced-topics/dockstore-cli/local-cromwell-config>` 
+
+:ref:`(back to top) <topCLIFAQ>`
+
+What environment do you test tools in?
+--------------------------------------
+
+Typically, we test running tools in Ubuntu Linux 16.04 LTS on VMs in
```
denis-yuen, line 123: I feel like this came from PCAWG originally, we don't really regularly test workflows on a standardized machine anymore, so this section could probably just be cut.

docs/dictionary.rst
```diff
@@ -890,6 +891,16 @@ parameter file
 
 
 
+.. _dict parent image:
+
+parent image
+------------
+    A :ref:`dict Docker image` which acts as the base upon which another Docker image is built. For example, including FROM ubuntu:18.04 in a :ref:`dict Dockerfile` means that the resulting image will include everything inside ubuntu:18.04, plus any changes made by other commands in the Dockerfile. Parent images are sometimes called base images, but strictly speaking a base image is different (see further reading).
```
denis-yuen, line 898: Ditto for 22.04

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
```diff
@@ -165,20 +175,29 @@ A new separate workflow/tool will be registered if the original name isn't inclu
...
+Any branches that do not have a .dockstore.yml file will not appear on your workflow/tool entry. If the branches you expect to see do indeed have properly formatted .dockstore.yml file, the culprit might be the name of the branches: The Dockstore GitHub App currently supports only the following special characters in branch names:
+* numerical digits
+* non-leading dashes
+* forward slashes
+* periods
+* underscores.
+
+Branches which contain other special characters, including alphabetical characters with accents, tildes, circumflexes, umlauts, or non-English letters such as ß and ø, cannot be parsed by the Dockstore GitHub App and will not show up. If you check the Dockstore GitHub App logs, you'll see these branches throw an error such as ``Reference refs/heads/Robert');-DROP-TABLE-Students; is not of the valid form``.
```
denis-yuen, line 195: :raised_eyebrow:

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
denis-yuen: e.g. our tooltester platform for metrics is now built on AWS AGC where each workflows presumably requests its own job sizes independently from AWS batch

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
```diff
@@ -17,6 +19,8 @@ FTP, S3, and GCS. As of Release 1.12, the Dockstore CLI has support for running
 client <https://docs.icgc.org/download/guide/#score-client-usage>`__. Please see `file provisioning plugins <https://github.com/dockstore/dockstore-cli/tree/master/dockstore-file-plugin-parent>`__
 for more information on these two file transfer sources.
 
+:ref:`(back to top) <topCLIFAQ>`
```
coverbeck, line 22: It seems like this PR also has commits for #247 included in it. It would be good to merge that one so we can see what's really different.

docs/faq.rst
```diff
@@ -297,42 +310,48 @@ Do you have tips on creating Dockerfiles?
...
+-  use a well-known and secure :ref:`dict parent image` such as official `debian <https://hub.docker.com/_/debian>`__ or `Python <https://hub.docker.com/_/python>`__ images
 -  try to keep your Docker images small
+   -  however, do not use alpine images, or other images that lack bash, as your parent image unless you will be installing bash in the Dockerfile 
+  
+
+:ref:`(back to top) <topFAQ>`
+
+How should I handle large reference files when designing workflows and Dockerfiles?
+-----------------------------------------------------------------------------------
+Generally speaking, you can choose to either "package" reference files in your Docker image, or you can treat them as "inputs" so they can be stage outside and mounted into the running container. Both of these approaches will work on a variety of backends, but there are some backends in which one option might be better than the other.
```
coverbeck, line 322: stage -> staged

docs/faq.rst
```diff
@@ -297,42 +310,48 @@ Do you have tips on creating Dockerfiles?
...
+
 
+      - if you are running on a backend that can mount files into your containers quickly, such as a local installation of miniwdl, and your input files are available without having to download them, is probably best to treat reference files as an input
```
coverbeck, line 325: `treat reference files as inputs`, IMO, although you can argue input is plural

docs/faq.rst
```diff
@@ -297,42 +310,48 @@ Do you have tips on creating Dockerfiles?
...
+      - if you are running your workflow on a backend that needs to download input files (such as Terra), and you need to put those reference files into many instances of a Docker container (such as if you are using scattered tasks), it is probably best to bake your reference files into the Dockerfile so they are immediately available in the Docker image
```
coverbeck, line 326: Hmm, this contradicts the advice  we give on [FAIR best practices](https://docs.dockstore.org/en/stable/advanced-topics/best-practices/best-practices-secure-fair-workflows.html?highlight=fair#image-container-best-practices), where we say "keep images light" and "don't include test data in the image".

I suggest dropping this section entirely, for now, until we sort that out.

You know a ton more about writing workflows than I do, and so this might be the right approach, but if so, then we should update the best FAIR practices as well.

docs/getting-started/github-apps/github-apps-troubleshooting-tips.rst
coverbeck: I think this whole section can be deleted. See dockstore/dockstore#5354

docs/faq.rst
aofarrel: ## short answer
Yeah, I tried to balance FAIR plus my own experience and it's a bit of a complicated topic. It can be argued that including a reference is more FAIR because it is more reproducible, especially with certain types of tooling. I'm in favor of dropping all references to image size in the short term, if that's what you mean?

## long answer (optional reading, sorry, I'm still verbose)
I came to the conclusion that sometimes it is best to include the reference thanks to working with a tuberculosis toolkit called [clockwork](https://github.com/iqbal-lab-org/clockwork). clockwork has a function ([which invokes this script](https://github.com/iqbal-lab-org/clockwork/blob/master/scripts/download_tb_reference_files.pl)) dedicated to downloading and preparing reference genomes, resulting in the creation of three reference genome folders -- one of which is about 15 GB in size because it contains human, Epstein-Barr, HIV, etc for decontamination purposes plus a ton of metadata. The other tasks expect those reference genome folders to be in the workdir, which is fine when you're working with the toolkit in an interactive way, but if you're trying to make this slightly modular as WDL tasks, you now have to pass around 15 GB of reference files around as inputs. clockwork doesn't have the ability to take in inputs as files, instead expecting them in the workdir, which isn't where WDL executors tend to localize files, so you need to also copy them [from an input folder with potentially odd permissions](https://miniwdl.readthedocs.io/en/latest/runner_advanced.html#read-only-input-files) into the workdir. So, those task-level inputs for the reference files are accessible to users, which means a user who has the reference files lying around could put them in manually... but having those inputs available increases the chance that a user may put the wrong thing in there, since explaining what the reference folder needs to contain can't be done very well in a short docstring. Also, WDL 1.0 does not accept folders as inputs or outputs, so you have to tarball them too, and all subsequent tasks need to include a untar command. [Terra is not very good at getting files in and out of Docker containers quickly](https://support.terra.bio/hc/en-us/community/posts/10732090690843-Switch-Localization-Delocalization-From-gsutil-to-gcloud-storage), so you're wasting compute time and resources on passing that tarball around. If you want to compress the tarball to make things faster, you're going to want to use [pigz](https://zlib.net/pigz/) because gzip is orders of magnitude slower on multi-GB files, but then you have to roll your own Docker image because [clockwork's provided Docker image](https://github.com/iqbal-lab-org/clockwork/pkgs/container/clockwork) doesn't come with pigz. [And if you're rolling your own Docker image, you may as well just include the reference files](https://github.com/aofarrel/clockwork-wdl/blob/main/build_dockers.sh). It certainly is possible to make passing in reference files work -- I did so until [2.9.0 of clockwork-wdl](https://github.com/aofarrel/clockwork-wdl/releases). But I realized including the references was easier for users (less inputs to keep track of) and less likely to break.

...of course, I might be the edge case here. I do know plenty of other tools expect reference files in specific places, so I think there is still a case to be made for including the references in the docker image for other tasks, but I don't want to overextrapolate here. 

(It should also be noted the Terra use case is a bit of a rock and a hard place situation -- Terra is sluggish at passing around 15 GB tarballs, but it doesn't cache Docker images either. That means **every** instance of a scattered task either has to pass in 15 GB of files via a slow gsutil command, or pull a 15 GB Docker image from Docker Hub. I suspect the Docker option is more cost-effective for the user but neither option is particularly good, especially considering Docker Hub sometimes fumbles the `docker pull` which crashes the entire workflow.)

coverbeck: > Yeah, I tried to balance FAIR plus my own experience and it's a bit of a complicated topic. It can be argued that including a reference is more FAIR because it is more reproducible, especially with certain types of tooling. I'm in favor of dropping all references to image size in the short term, if that's what you mean?

What I meant is the doc should have the same recommendation everywhere. With this PR, the doc will:

* On the FAIR page, say "don't include data in Docker images", period.
* This page will say "sometimes include data in the Docker images"

I think we should have the same message throughout the doc, and until we do, we shouldn't introduce divergences, even if in the end we end up with your recommendation.

aofarrel: Agreed completely and I'll be deleting the "sometimes include data" from this page as brought up for that reason. Thanks for flagging this, I had forgotten that our FAIR page had that recommendation. It's important docs don't contradict.

docs/advanced-topics/dockstore-cli/dockstore-cli-faq.rst
aofarrel: Removed the question entirely and replaced it with a question about launching workflows in general in the general FAQ

aofarrel:
New changes since review:
* Add "how do I use workflows" to general FAQ (we don't seem to have a generalized "launch with cloud partners" page so we're kind of counting on users having a platform in mind and clicking it in the sidebar)
* Remove regex question thanks to https://github.com/dockstore/dockstore/pull/5354
* Remove "what environment do you test on" question
* Large reference files question still acknowledges you can put inputs into the Dockerfile, but recommends the inputs method -- we can revisit that later, but it's important to not contradict the FAIR page
* Stage/Staging typo fixed
* Ubuntu 22.04 instead of 18.04

PR #254, opened 2023-12-18
Update Getting Started with Galaxy documentation

coverbeck:
The ticket's goal was to have a sample workflow in GitHub with a .dockstore.yml that users could fork, but after going over the Galaxy doc and brief exchange with a Galaxy developer, I think it makes more sense to embrace the Galaxy way of writing workflows, through their UI, and not try to steer Galaxy users towards the CWL/WDL way of writing workflows.

The developer pointed out a great IWC link that talks about best practices in Galaxy workflows, and includes a reference to a command-line tool that generates a .dockstore.yml. So I've tended to rely more on linking that writing our own doc, that risks getting out of sync, and won't be as knowledgeable as doc written by the Galaxy team.

Also fixed links that had broken over time (build failed otherwise).

dockstore/dockstore#3713
DOCK-1520


coverbeck:
Will run add-discourse-topic action (and build will pass) after/if PR is approved.

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -28,13 +28,15 @@ Using Custom Galaxy in the Cloud with Terra, and AnVIL
...
-Unlike WDL and CWL, Galaxy workflows in the near term are created and modified from the Galaxy workflow editor (GUI), instead of a text editor.
-
 Create a basic Galaxy workflow
 ------------------------------
 
+Unlike WDL and CWL, Galaxy workflows are currently created and modified from the Galaxy workflow editor (GUI), instead of a text editor.
+
 If you would like to create and run your own workflow in Galaxy, here  is a tutorial for `Creating, Editing, Importing Galaxy Workflows <https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/workflow-editor/tutorial.html>`__ from the Galaxy Training Network.
 
+In addition, the Galaxy Intergalactic Workflow Commission (IWC) team has an excellent tutorial for creating best practices workflows for IWC `here <https://github.com/galaxyproject/iwc/blob/main/workflows/README.md>`__. Although somewhat specific to the IWC directory structure, it discusses how to add the correct metadata, generate tests, and lint your workflow. It also shows how you can generate a ``.dockstore.yml`` using the `Planemo <https://planemo.readthedocs.io/en/latest/index.html>`__ command-line utilities.
```
svonworl, line 38: Might be good to develop "Although somewhat specific to the IWC directory structure" to explain to the user why we've noted it.

svonworl (approved):

denis-yuen:
> So I've tended to rely more on linking that writing our own doc, that risks getting out of sync, and won't be as knowledgeable as doc written by the Galaxy team.

Think this makes sense

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -28,13 +28,15 @@ Using Custom Galaxy in the Cloud with Terra, and AnVIL
 ------------------------------------------------------------------------
 Galaxy is also integrated into the Terra cloud workspace. This provides users the ability to readily create a custom Galaxy environment scaled to their needs or with additional tools installed. Additionally, the Terra workspace is able to access a rich data corpus hosted in the cloud, including many controlled access datasets, thus allowing users to conduct analyses with this data in a compliance-based environment.
 
-Unlike WDL and CWL, Galaxy workflows in the near term are created and modified from the Galaxy workflow editor (GUI), instead of a text editor.
-
 Create a basic Galaxy workflow
 ------------------------------
 
+Unlike WDL and CWL, Galaxy workflows are currently created and modified from the Galaxy workflow editor (GUI), instead of a text editor.
```
denis-yuen, line 34: Sentence may be a little weird if the user starts with Galaxy and doesn't know CWL or WDL.

denis-yuen (commented):
Does this branch make sense?
Looks like yes, there is no `release/1.15` branch yet

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -53,6 +55,13 @@ outputs, and Galaxy Tool Shed dependencies for your workflow.
 
    Download
 
+If you've installed `Planemo <https://planemo.readthedocs.io/en/latest/index.html>`__, generate a ``.dockstore.yml`` file according to the `IWC instructions <https://github.com/galaxyproject/iwc/blob/main/workflows/README.md>`__, by navigating to the directory containing your workflow and run:
+
+::
+
+$ planemo dockstore_init .
+
+The ``.dockstore.yml`` is used in conjunction with the :doc:`Dockstore GitHub App <github-apps/github-apps-landing-page>`. For other workflow languages, the ``.dockstore.yml`` is created in a text editor; for Galaxy you have the option of generating with a command-line tool, and it's recommended you do so.
```
kathy-t, line 64: This sentence feels like it should come before `planemo dockstore_init` since it explains why you want to generate one using planemo.

kathy-t (approved):

docs/dictionary.rst
```diff
@@ -1,1116 +0,0 @@
-.. DO NOT EDIT THIS FILE. This file is autogenerated from glossary_generator.py, update that instead.
```
denis-yuen, line 1: Seems odd, did the auto-generation fail? Was this not meant to be checked in?
File has a couple years of history, so thinking the former

denis-yuen: Hmmm, I see the commit comment: 
```
It's a generated file. I did not know this, kept editing it to
make a change, only to see my change get overwritten. Delete
it to avoid future confusion.
```

Is it worth checking that the dictionary is properly auto-generated/displayed by readthedocs without this file checked in?

coverbeck: My first commit to the file was done because the generator had changed it. After pushing it, I belatedly realized that would be confusing in the future, so I did a second commit to just delete the file.

> Is it worth checking that the dictionary is properly auto-generated/displayed by readthedocs without this file checked in?

I'll take a look.

docs/getting-started/getting-started-with-galaxy.rst
```diff
@@ -65,12 +67,23 @@ are not familiar with GitHub you can use this
 `tutorial <https://guides.github.com/activities/hello-world/>`__ to set up
 an account and repository.
 
+Creating a .dockstore.yml
+-------------------------
+
+You will need a ``.dockstore.yml`` file. It is used in conjunction with the :doc:`Dockstore GitHub App <github-apps/github-apps-landing-page>` to integrate your workflow with Dockstore. For other workflow languages, the ``.dockstore.yml`` is created in a text editor; for Galaxy you have the option of generating with a command-line tool, and it's recommended you do so.
```
svonworl, line 73: Borderline nit, but we might be able to condense the first two sentences into one that is more plain English ("in conjunction with" is a bit stilted, imho) and explains the purpose of `.dockstore.yml` and how our App uses it process the repo.

svonworl (approved):

docs/getting-started/getting-started-with-galaxy.rst
coverbeck: Stilted, dost thou sayest? :) Fair point, I'll try to come up with something better.

The link in the second sentence is to our main documentation on how GitHub and .dockstore.yml work. I didn't want to duplicate that nor rewrite it here. On the other hand, we are asking the user to generate the file here, without having read that portion yet, so I'll try to come up with a little more detail.

coverbeck:
Made the following changes:

1. Added a check in the config.yml to ensure the dictionary.rst is generated
2. Reworded and moved the section about generating a .dockstore.yml to another section where I should have had it in the first place.
3. Added a reference from the generic .dockstore.yml file creation back to Galaxy .dockstore.yml generation.

kathy-t (approved):

denis-yuen (approved):

svonworl (approved):

PR #256, opened 2024-01-19
Fix build by pinning sphinxcontrib packages

svonworl:
**Description**
When I cloned this repo, created and activated a new venv, and attempted to build within, the build failed at the `make html` step, with the error:
```
(env) bash-3.2$ make html
make glossary
echo "\033[1mGenerating dictionary.rst...\033[0m"
Generating dictionary.rst...
python3 _attic/glossary_generator.py
Running Sphinx v4.4.0
loading translations [en]... done

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
make: *** [html] Error 2
(env) bash-3.2$
```

Indeed, the more recent releases of the various `sphinxcontrib` packages do require Sphinx 5.0 (or higher).  This new requirement is documented in the `CHANGES` file for each package, for example:
https://github.com/sphinx-doc/sphinxcontrib-applehelp/blob/c86aa1d216a3605b5781bb3f90c8f399e6cf616f/CHANGES#L22

So, in `requirements.txt`, I pinned each package to the version preceding the Sphinx 5.0 requirement.  Some of these require Python 3.9.

Assuming that we don't want to upgrade to a newer version of Sphinx, is this the Pythonic way to deal with this issue?

(Thanks to Charles for help diagnosing the problem.)

**Issue**
None.


denis-yuen:
> Assuming that we don't want to upgrade to a newer version of Sphinx, is this the Pythonic way to deal with this issue?

I don't really have an objection to upgrading to Sphinx. 
That said, pinning everything is basically what we do with Maven so it would be an improvement on letting things slide (at least IMHO)

denis-yuen (approved):

svonworl:
> > Assuming that we don't want to upgrade to a newer version of Sphinx, is this the Pythonic way to deal with this issue?
> 
> I don't really have an objection to upgrading to Sphinx. That said, pinning everything is basically what we do with Maven so it would be an improvement on letting things slide (at least IMHO)

Upgrading to Sphinx 5.0 triggers some other version changes, and a different build issue.  Which is solvable, but it makes me wonder what else has silently broken.  Probably something to do after a release, so we have time to observe and find problems.

requirements.txt
```diff
@@ -4,3 +4,9 @@ sphinx_rtd_theme==1.0.0
 git+https://github.com/pdavide/sphinxcontrib-discourse
 attrs==21.4.0
 glossarpy==0.1.2
+sphinxcontrib-applehelp==1.0.4
```
coverbeck, line 7: Disclaimer, not a Python expert, but I was looking -- shouldn't we do a `pip freeze` to pin all dependencies? This will fix the immediate problem, but without the freeze, couldn't a variant of the problem might creep back in?

coverbeck (approved):

david4096 (approved):

requirements.txt
david4096: pip freeze just generates a requirements file based on the currently installed packages in a given environment. Agreed with the above that we should just upgrade when its expedient (and then pin the packages to a known working version of each).

coverbeck: Presumably we didn't do a `pip freeze` before, or this wouldn't have come up?

svonworl: I committed a new `requirements.txt` that freezes all dependencies.  The process was to create a clean virtual env, run `pip install -r requirements.txt`, confirm a successful build, and generate a new `requirements.txt` via `pip freeze`.

kathy-t (approved):

PR #259, opened 2024-01-22
Add information about .dockstore.yml "topic" field

svonworl:
**Description**
This PR modifies our documentation to explain the new `.dockstore.yml` "topic" field.  Specifically, it:
1. Adds the "topic" line to the example/template `.dockstore.yml`s.  Even the short `.dockstore.yml` examples now include the `topic` field, because IMHO, we want users to include a "topic", because it helps other users process the search results and understand an entry's purpose/function/etc.
2. Add a "topic" entry to the Dockstore dictionary.

Going into the ticket, I expected to modify the existing documentation that explained the general concept of an entry's topic.  But, I couldn't find any.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2480


docs/assets/templates/notebooks/template-big.dockstore.yml
```diff
@@ -49,7 +49,13 @@ notebooks:
         email: <String>
         role: <String>
         affiliation: <String>
-    
+
+    # A short text description of the notebook, 140 characters or less in length.
+    # Notebook-wide setting that will affect ALL branches/tags; only set this as needed in a main branch.
+    # If no topic has been specified, Dockstore will use the repository's "About" description as the topic.
```
denis-yuen, line 55: May need to check the code, but this may only work for GitHub (for example)

docs/assets/templates/services/template-big.dockstore.yml
```diff
@@ -18,6 +18,12 @@ service:
       role:
       affiliation:
 
+  # A short text description of the service, 140 characters or less in length.
+  # Service-wide setting that will affect ALL branches/tags; only set this as needed in a main branch.
+  # If no topic has been specified, Dockstore will use the repository's "About" description as the topic.
+  # Set this field to the empty string to reset the topic to the repository's "About" description.
+  topic:
+
```
denis-yuen, line 26: Should probably note upfront that this is optional (and not even necessarily recommended) a bit like the line below

denis-yuen: i.e. we recommend it more in the case of multiple workflows in a repo, not in general

denis-yuen (commented):
some elaboration may be helpful

docs/assets/templates/services/template-big.dockstore.yml
svonworl: As I created this PR, I debated the following question with myself: should we prefer the user to specify a `topic` in `.dockstore.yml` for a single-entry repo?  In the end, I concluded "yes", for the following reasons;
1. If we can get the user to include a topic, for example by highlighting the `topic` field in the `.dockstore.yml` examples, we don't have to rely on them setting the repo "About" description, which is easy to forget or just not do.
2. The repo [and thus, its "About" description] may contain more than just stuff for Dockstore.
3. The user might want their repo "About" description to be longer than the topic limit.  I did a quick test, the github "About" description can be up to 351 (!?) chars long.

However, I can understand arguments in the other direction.  Would like to hear what others think...

docs/assets/templates/notebooks/template-big.dockstore.yml
svonworl: Looks like it works for non-`.dockstore.yml` workflows and apptools, at least, see the `setTopicAutomatic` calls in `GitHubSourceCodeRepo`.  For notebooks and services, `.dockstore.yml` is the only registration method.

denis-yuen: I think that makes sense, mostly GitHub only was the main caveat I think

docs/getting-started/github-apps/migrating-workflows-to-github-apps.rst
```diff
@@ -34,12 +34,14 @@ As noted in our other documentation, create a .dockstore.yml file in the root di
...
         testParameterFiles:
             - /test/dockstore.cwl.json
+        topic: A short description of the workflow
 
 The information above was filled out using the following:
 
 - ``subclass`` is taken from the ``Descriptor Type``
 - ``primaryDescriptorPath`` is from ``Workflow Path``
 - ``testParameterFiles`` is from ``Test File Path``
+- ``topic`` is a short text description of the workflow, 140 characters or less.
```
kathy-t, line 44: 140 -> 150

kathy-t (approved):

coverbeck (approved):

docs/assets/templates/notebooks/template-big.dockstore.yml
svonworl: > I think that makes sense, mostly GitHub only was the main caveat I think

The above comments are in the long example templates for `.dockstore.yml`, which is GitHub only, so it's implicit.  We can add verbiage about it being GitHub only, but not sure it's worth the extra words.  (IMHO, one of the issues with our docs is that the ratio of information to words is too high...)

docs/assets/templates/services/template-big.dockstore.yml
svonworl: > If it's recommended, we should also note that 😁
> 
> I was mostly thinking in terms of the simplest case, a single workflow in the repo and them being too lazy to have a description in both the .dockstore.yml and the about field on GitHub. But I can be persuaded

This PR includes the `topic` field in most of the `.dockstore.yml` examples, and calls it out as "optional" in the long examples, so it feels like we're covering the two bases ("recommended" and "optional").  But if you don't think that's adequate, leave a comment and I'll try to figure something out.

docs/assets/templates/notebooks/template-big.dockstore.yml
denis-yuen: I don't know if that will always be the case (GitHub only), but fair enough

docs/assets/templates/services/template-big.dockstore.yml
denis-yuen: I do think the long example should say a few words about why we recommend it, partly due to the discussion above.

 i.e. it can be as simple at "you can add a description here if you don't have a description in the repo and especially if you have multiple workflows in a repo and want each one to have a unique description "

I think touching on the "why" is useful here

hyunnaye (approved):

docs/assets/templates/services/template-big.dockstore.yml
svonworl: > I do think the long example should say a few words about why we recommend it, partly due to the discussion above.
> 
> i.e. it can be as simple at "you can add a description here if you don't have a description in the repo and especially if you have multiple workflows in a repo and want each one to have a unique description "
> 
> I think touching on the "why" is useful here

I added a line to the comments in the "long" `.dockstore.yml` examples:
```
# Useful to specify unique topics for multiple notebooks in the same repository, or a topic that differs from the default.
```
I also reordered the lines to improve the flow.  Note that new line for the "service" example only includes the second part, because we don't support multiple services in the same repo.

denis-yuen (approved):

PR #262, opened 2024-01-24
SEAB-3518: added line regarding requester pays in the FAQ

hyunnaye:
**Description**
This PR adds a line in the Open Data section of the FAQ that mentions that while our facet is open and free, accessing "non open data" may be requester's pay. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3518

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -78,7 +78,9 @@ Note that tool and workflow versions that do not require any file input paramete
 You can find Open Data workflows and tools by selecting the Open Data facet on the `search <https://dockstore.org/search>`_ page.
 When viewing an individual tool or workflow, you can tell which versions are Open Data on the Versions tab via the Open column.
 
-While our Open Data facet is conservative and only considers freely accessible data, other data marked as `requester pays <https://cloud.google.com/storage/docs/requester-pays>`_ may be trivial to access given payment".
+.. note::
+    Dockstore does not consider requester pays data such as `Google Cloud Requester Pays <https://cloud.google.com/storage/docs/requester-pays>`_  and `Amazon S3 Requester Pays <https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html>`_ to be open data.
+    Even though their data may be open, our Open Data facet only show those that are open and free.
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -78,6 +78,8 @@ Note that tool and workflow versions that do not require any file input paramete
 You can find Open Data workflows and tools by selecting the Open Data facet on the `search <https://dockstore.org/search>`_ page.
 When viewing an individual tool or workflow, you can tell which versions are Open Data on the Versions tab via the Open column.
 
+While our Open Data facet is conservative and only considers freely accessible data, other data marked as `requester pays <https://cloud.google.com/storage/docs/requester-pays>`_ may be trivial to access given payment".
```
coverbeck, line 81: Suggest you do this as Note, so it sticks out a bit more.

The wording seems a little complex to me; I suggest something like Dockstore does not consider requester pays data to be open data, even though...

I believe linking to GS like that implies this is only Google Cloud thing; I suggest qualifying the link with "such as" and/or also having a link to the s3's requester pays.

svonworl (approved):

docs/faq.rst
```diff
@@ -78,6 +78,10 @@ Note that tool and workflow versions that do not require any file input paramete
 You can find Open Data workflows and tools by selecting the Open Data facet on the `search <https://dockstore.org/search>`_ page.
 When viewing an individual tool or workflow, you can tell which versions are Open Data on the Versions tab via the Open column.
 
+.. note::
+    Dockstore does not consider requester pays data such as `Google Cloud Requester Pays <https://cloud.google.com/storage/docs/requester-pays>`_  and `Amazon S3 Requester Pays <https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html>`_ to be open data.
+    Even though their data may be open, our Open Data facet only show those that are open and free.
```
coverbeck, line 83: "our Open Data facet only show those that are open and free" - should be `shows`.

coverbeck (approved):

denis-yuen (approved):

PR #264, opened 2024-01-26
SEAB-6957: Documentation for delete entry feature

hyunnaye:
**Description**
This PR creates a new documentation page with the steps to delete an entry.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6057


svonworl (approved):
This looks good!  We did recently change the UI to remove the "Is not a checked or checker workflow" text from the dialog for notebooks, so you might want to go to qa and capture another image.

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
...
+-  Has never been published,
+-  Was created on/after April 1, 2022, and
+-  Is not a checked or checker workflow.
+
+
+Deleting an Entry
+------------------
+
+Navigate to ``/my-tools``, ``/my-workflows``, ``/my-notebooks``, or ``/my-services`` by going to My Dashboard then selecting the desired option in the left sidebar.
+In the sidebar accordion, find the github repository that your entry is in and click the entry you would like to delete.
```
kathy-t, line 27: github -> GitHub

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
+Deleting an entry on Dockstore
```
kathy-t, line 1: Nit: entry -> Entry because the other side bar titles are capitalized

image

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
...
+-  Deleting an entry
+-  Uninstalling the Dockstore GitHub App from the source GitHub repo
+
+This tutorial walks through the process of deleting entries from Dockstore.
+This action is permanent and after you delete the workflow, it will no longer appear on Dockstore.
+
+Requirements for Deletion
+-------------------------
+
+An entry can only be deleted if it satisfies all the following:
```
kathy-t, line 16: May be worth mentioning that if they don't meet the requirements to delete, they can archive it

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
+Deleting an entry on Dockstore
+===================================
+
+Tutorial Goals
+--------------
+
+-  Deleting an entry
+-  Uninstalling the Dockstore GitHub App from the source GitHub repo
```
kathy-t, line 8: Pondering if a more generic sentence should be used like "Ensuring that the entry doesn't re-appear" since below you mention that you could also remove the entry from the .dockstore.yml or delete the .dockstore.yml

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
...
+
+.. image:: /assets/images/docs/notebook-showing-delete-button.png
+   :width: 50 %
+
+Read the dialog message and if you are sure you would like to delete, click the Delete this workflow/tool/notebook/service button
+
+.. image:: /assets/images/docs/delete-notebook-dialog.png
+   :width: 50 %
+
+After you delete the entry, you must uninstall the Dockstore GitHub App from the source GitHub repo, or edit/remove the .dockstore.yml file so that it no longer describes the deleted entry.
```
kathy-t, line 42: Seems like this tutorial is for deleting GitHub App entries because this step isn't applicable to legacy/hosted tools and workflows. Maybe we should move it to the Dockstore GitHub App section to avoid possible confusion? https://docs.dockstore.org/en/stable/getting-started/github-apps/github-apps-landing-page.html

docs/assets/images/docs/sidebar-accordian-notebook-select.png
kathy-t, line 1: We also removed the Refresh Organization button because it's not applicable to notebooks so this screenshot should be updated

docs/getting-started/delete-on-dockstore.rst
denis-yuen: Think this is worth it's own article, could also add this to the list of requirements about though

docs/getting-started/delete-on-dockstore.rst
coverbeck: I still see `github`. Maybe you just haven't pushed yet, but flagging it just in case.

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,44 @@
...
+===================================
+
+Tutorial Goals
+--------------
+
+-  Deleting an entry
+-  Uninstalling the Dockstore GitHub App from the source GitHub repo
+
+This tutorial walks through the process of deleting entries from Dockstore.
+This action is permanent and after you delete the workflow, it will no longer appear on Dockstore.
```
coverbeck, line 11: Nit: You're following the dialog text, where I should have noticed it earlier, but it seems little dissonant to on the one hand say it's permanent and will no longer appear on Dockstore, and then there is a section below on how to ensure is doesn't appear again -- that's not permanent. :)

Not sure I have an easy rephrasing in mind, so just something to ponder, and if no other idea comes, ok to leave.

coverbeck (approved):

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,45 @@
...
+-------------------------
+
+An entry can only be deleted if it satisfies all the following:
+
+-  Registered via GitHub Apps,
+-  Has never been published,
+-  Was created on/after April 1, 2022, and
+-  Is not a checked or checker workflow.
+
+If your entry does not meet the requirements above, you can archive it using these steps.
```
hyunnaye, line 22: I'll link the documentation page for archiving in my next PR for https://ucsc-cgl.atlassian.net/browse/SEAB-5979

docs/getting-started/delete-on-dockstore.rst
hyunnaye: Kept line to keep consistent with dialog

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -0,0 +1,46 @@
...
+
+This tutorial walks through the process of deleting entries from Dockstore.
+This action is permanent and after you delete the workflow, it will no longer appear on Dockstore.
+
+Requirements for Deletion
+-------------------------
+
+An entry can only be deleted if it satisfies all the following:
+
+-  Registered via GitHub Apps,
```
denis-yuen, line 18: Actually, I may have misread the conversation below. 
I think deletion in general works for all workflows (fulfilling the requirements below)

It's just the note on like 44 which applies only to GitHub apps and is probably fine to leave here (since we may eventually want a Bitbucket app, etc.)

denis-yuen (commented):
I think I mis-spoke. 
I don't think deletion here applies only to GitHub app workflows. 
It's just that you only need to remove the app for workflows of that kind when deleting

docs/getting-started/delete-on-dockstore.rst
```diff
@@ -1,11 +1,11 @@
...
 
 Tutorial Goals
 --------------
 
 -  Deleting an entry
--  Uninstalling the Dockstore GitHub App from the source GitHub repo
+-  Ensuring that the entry does not re-appear on Dockstore
 
 This tutorial walks through the process of deleting entries from Dockstore.
 This action is permanent and after you delete the workflow, it will no longer appear on Dockstore.
```
kathy-t, line 11: Suggesting saying "entry" in "after you delete the workflow" instead of "workflow" to be more generic

kathy-t (approved):

denis-yuen (approved):

PR #268, opened 2024-02-07
SEAB-5979: Documentation for the Archive feature

hyunnaye:
**Description**
This PR adds documentation for the new Archive feature.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5979


docs/getting-started/archive-on-dockstore.rst
```diff
@@ -0,0 +1,68 @@
...
+- Marked as "Archived".
+- Read-only, except to unarchive. GitHub updates will be ignored.
+- Ranked lower in Dockstore Search.
+- Displayed in the "Archived" section of your Dockstore dashboard.
+
+Archiving an Entry
+-------------------
+
+Navigate to ``/my-tools``, ``/my-workflows``, ``/my-notebooks``, or ``/my-services`` by going to My Dashboard then selecting the desired option in the left sidebar.
+In the sidebar accordion, find the GitHub repository that your entry is in and click the entry you would like to archive.
```
coverbeck, line 30: I don't think Archiving is a GitHub-only thing; it works for hosted workflows, and should work for Gitlab and Bitbucket as well.

docs/getting-started/archive-on-dockstore.rst
```diff
@@ -0,0 +1,68 @@
...
+This tutorial walks through the process of archiving entries from Dockstore.
+
+This action will make your entry read-only and less prominent in the Dockstore user interface.
+You can unarchive your entry back to normal at any time.
+
+
+Specific Changes When Archived
+-------------------------------
+
+After you archive the Dockstore workflow, it will be:
```
coverbeck, line 19: Here you talk about archiving a workflow, then below, e.g., line 53, you talk about archiving a tool, workflow, notebook, or service. Either is good, but it should be consistent.

docs/getting-started/archive-on-dockstore.rst
```diff
@@ -0,0 +1,68 @@
+Archiving an Entry on Dockstore
+===================================
+
+Tutorial Goals
+--------------
+
+-  Archiving an entry
+-  Unarchiving an entry
+
+This tutorial walks through the process of archiving entries from Dockstore.
```
coverbeck, line 10: The tutorial archives a single entry, not entries.

docs/getting-started/archive-on-dockstore.rst
svonworl: > Here you talk about archiving a workflow, then below, e.g., line 53, you talk about archiving a tool, workflow, notebook, or service. Either is good, but it should be consistent.

IMHO, would be good if we use terminology that covers all entry types, because you can delete any type of entry.  "Tool, workflow, notebook, or service" would be ok, although it's long, and not scalable: it'll get even longer when we add the next Entry type.  Maybe it's time to reconsider our aversion to "entry" (or come up with a more palatable equivalent).

svonworl (approved):

denis-yuen (commented):
With comments above

denis-yuen (approved):

docs/getting-started/archive-on-dockstore.rst
```diff
@@ -0,0 +1,68 @@
...
+You will now see a banner that shows that your entry is archived and read-only.
+
+.. image:: /assets/images/docs/archived-workflow-banner.png
+   :width: 50 %
+
+Unarchiving an Entry
+---------------------
+
+Navigate to ``/my-tools``, ``/my-workflows``, ``/my-notebooks``, or ``/my-services`` by going to My Dashboard then selecting the desired option in the left sidebar.
+In the sidebar accordion, find the GitHub repository that your entry is in and click the entry you would like to unarchive in the Archived section.
```
coverbeck, line 54: This still has `GitHub repository`.

docs/getting-started/archive-on-dockstore.rst
coverbeck: Fair point in general, but in this case I was thinking of it as a tutorial, where you usually do one thing (and there is a screenshot of a workflow, or is a tool tool), so one option would have been at the beginning to say something like "this tutorial applies to workflows, tools, services, and notebooks, but here we'll be using a workflow as an example).

Nayeon, no need to change anything, just elaborating on my comment.

coverbeck (approved):

kathy-t (approved):

PR #271, opened 2024-02-15
DOCK-2466: Improve FAQ discussion of allowing OAuth access

svonworl:
**Description**
This PR modifies an existing section of the FAQ to clarify a bit of GitHub UI that seemed to scare/confuse a user:

image

This message, with the frightening red X and "private access", appears when the Dockstore application is not granted access to the organization via OAuth.

The webservice is granted access via two mechanisms: OAuth and the GitHub App.  As far as I can tell, allowing the GitHub App allows the webhooks to work, and thus the registration mechanisms built on top of them.  Granting OAuth access allows all of the non-webhook endpoints that access GitHub information about the user to work, enabling features such as manual registration and "Discover Existing".  It's possible to grant access to one and not the other, thus enabling partial functionality.

In the end, I decided to encourage the admin to grant OAuth access, so that our features are more likely to work 100%, and not even discuss the "halfway" configuration, which seemed like it would just confuse the issue, and cause other support problems down the line.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2466
dockstore/dockstore#5676


docs/faq.rst
```diff
@@ -193,7 +193,8 @@ Why are my workflows from an organization I belong to not visible?
 Organizations have the ability to restrict access to the API for third
 party applications. GitHub provides a
 `tutorial <https://help.github.com/en/articles/enabling-oauth-app-access-restrictions-for-your-organization/>`__
-on how to add these restrictions to your organizations.
+on how to add these restrictions to your organizations.  For recently-created GitHub organizations,
```
denis-yuen, line 196: Wasn't aware of this, do you have a handy link to read up on this?

docs/faq.rst
```diff
@@ -203,6 +204,18 @@ itself. GitHub provides a
...
 
+As you navigate GitHub's administrative interface, you may see a
+message indicating that you have denied Dockstore access
+to your organization:
+
+.. image:: /assets/images/docs/dockstore-oauth-access-denied.png
+
+To grant access to the Dockstore application, click the *Grant access* button.
+
+If you don't grant access, some Dockstore features, such as manual workflow
```
denis-yuen, line 215: Instead of "will not" maybe "may not" at least for workflow discovery. 
As per the user in the original discussion and as per our discussion, it seems that Dockstore may just work "enough" for regular github app use, publishing, etc.

docs/faq.rst
svonworl: > Wasn't aware of this, do you have a handy link to read up on this?

See the first paragraph in body-sized type:
https://docs.github.com/en/organizations/managing-oauth-access-to-your-organizations-data/enabling-oauth-app-access-restrictions-for-your-organization
I tested this a long time ago, so I did it again just now to make sure. Indeed, in a fresh, new, untouched dummy GitHub org, indeed, the restrictions are enabled by default:
image

denis-yuen: Annoying change, but thanks!

coverbeck (approved):

docs/faq.rst
svonworl: > Instead of "will not" maybe "may not" at least for workflow discovery. As per the user in the original discussion and as per our discussion, it seems that Dockstore may just work "enough" for regular github app use, publishing, etc.

I can soften it, but maybe also refer specifically to the "Discover Existing" functionality, rather than saying "workflow discovery", to avoid any confusion with being able to search or publish.

svonworl: Food for thought: yesterday, Charles and I were talking about the idea of authorizing everything as our GitHub App, which you could do with the `installationId`, I think, same way the webhook endpoints do.  If we recorded App installs, I think the `installationId` arrives with that.

kathy-t (approved):

docs/faq.rst
denis-yuen: I think it would simplify things to remove OAuth.
I did an experiment in https://ucsc-cgl.atlassian.net/browse/SEAB-5915 and it seemed like most things just work using the GitHub app only

denis-yuen (approved):

PR #273, opened 2024-02-29
DOCK-2468: Better document yaml array syntax

svonworl:
**Description**
This PR modifies our documentation so that users are more likely to understand that they need to set the `.dockstore.yml` `<String Array>` fields to a proper YAML array.

It's important to note that users will always ignore our docs and wing it.  The "final boss" solution (not implemented here) is to avoid having to create error-prone documents to use our product.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2468
dockstore/dockstore#5683


docs/assets/templates/notebooks/template-big.dockstore.yml
```diff
@@ -33,7 +33,7 @@ notebooks:
     publish: <Boolean>
 
     # An optional array of absolute paths to other files in the Git repository that should be read and included with the notebook.
-    otherFiles: <String Array>
+    otherFiles: <String Array>  # Must be a YAML array of strings
```
denis-yuen, line 36: Honestly, I'd often have to look up array syntax when switching languages. 
I think a template example is more intuitive since a user can copy-paste and replace
i.e.
```
otherFiles:
  - /exampleFile1
  - /exampleFile2
```

docs/assets/templates/notebooks/template-big.dockstore.yml
```diff
@@ -69,5 +69,5 @@ notebooks:
     # https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)
     # or RegEx when the string is surrounded by / (Ex: `/develop/`, `/mynotebook\/.*/`).
     filters:
-      branches: <String Array>
-      tags: <String Array>
+      branches: <String Array>  # Must be a YAML array of strings
```
denis-yuen, line 72: Ditto, a bunch more below

docs/assets/templates/yaml-array-advice.rst
```diff
@@ -0,0 +1,12 @@
+When setting a field that accepts a ``<String Array>``, be sure to specify the list of strings in YAML format.  For example, to set the ``testParameterFiles`` field, you can use either the "one per line" YAML array format:
+
+::
+
+  testParameterFiles:
+    - /path/to/test.json
```
denis-yuen, line 6: Should specify more than one line in the example

denis-yuen (changes requested):

docs/assets/templates/notebooks/template-small.dockstore.yml
```diff
@@ -4,12 +4,10 @@ notebooks:
     language: <PYTHON | R | JAVASCRIPT | SCALA | JULIA | OTHER>
     path: <String>
     kernel: <String>
-    otherFiles: <String Array>
-    testParameterFiles: <String Array>
+    otherFiles: <String Array>  # Must be a YAML array of strings
+    testParameterFiles: <String Array>  # Must be a YAML array of strings
```
kathy-t, line 8: Do notebooks have test parameter files? The big template doesn't show it

svonworl: > Do notebooks have test parameter files? The big template doesn't show it

At present, we allow them to be specified, although we're not sure what they mean yet.

docs/assets/templates/notebooks/template-big.dockstore.yml
svonworl: We can do something like that, but these `<String Arrays>` appear in the "generic" template, we have "filled out" examples below on the same page.  It wouldn't make sense to have actual paths in the generic example.  So, maybe something like:
```
otherFiles:
   - <String>
   - <String>
```
And if it's as above, we might be able to delete the advice about array formats that I added, because we're demonstrating it above.

docs/assets/templates/tools/template-small.dockstore.yml
```diff
@@ -2,7 +2,7 @@ version: 1.2
 tools:
   - subclass: CWL
     primaryDescriptorPath: <String>
-    testParameterFiles: <String Array>
+    testParameterFiles: <String Array>  # Must be a YAML array of strings
     authors:
       - name: <String>
```
coverbeck, line 7: Here's a precedent for having an actual array, although this one's not a simple list of strings.

docs/assets/templates/notebooks/template-big.dockstore.yml
coverbeck:
```
otherFiles:
   - <String>
   - <String>
```

This is also how we do it with authors (see my other comment). I like this proposal.

svonworl:
Ok, I modified the PR so, essentially, it changes instances of:
```
    field: <String Array>
```
to
```
    field:
       - <String>
       - <String>
```

kathy-t (approved):

denis-yuen (approved):

coverbeck (approved):

PR #274, opened 2024-03-04
Generate dictionary.rst in readthedocs build

kathy-t:
**Description**
While reviewing the notebooks documentation for https://ucsc-cgl.atlassian.net/browse/SEAB-5124, I noticed that all references to the dictionary were broken, displaying `dict <dictionary term>` such as `dict TRS API` below.

image

There are other instances of it [here](https://docs.dockstore.org/en/stable/search.html?q=dict&check_keywords=yes&area=default).

Was able to replicate it locally by removing the line `make glossary` from the [Makefile](https://github.com/dockstore/dockstore-documentation/blob/e9740080eb182a91cb643e8657b3d6b8e998081e/docs/Makefile#L23), deleting the entire `docs/_build` directory to remove the existing dictionary.rst that was previously generated, and re-running `make html`.

It seems like [the build process on readthedocs](https://readthedocs.org/projects/dockstore/builds/23065004/) does not generate this dictionary.rst, resulting in this error in the final build step:

`/home/docs/checkouts/readthedocs.org/user_builds/dockstore/checkouts/develop/docs/index.rst:16: WARNING: toctree contains reference to nonexisting document 'dictionary'`

The dictionary.rst file was deleted in https://github.com/dockstore/dockstore-documentation/pull/254. Instead of adding it back, this PR generates the dictionary.rst file in the readthedocs build which is already done by the local build via the Makefile.

<s>This fix is untested because I can't trigger the readthedocs build. If someone with permissions is able to activate this branch on readthedocs so I can verify the build, it would be appreciated (@denis-yuen maybe?)</s>

Can view the fixed build at https://readthedocs.org/projects/dockstore/builds/23639716/ and the docs at https://docs.dockstore.org/en/feature-check-in-dictionary/getting-started/getting-started-with-notebooks.html

**Issue**
n/a found while verifying notebooks documentation


denis-yuen:
> This fix is untested because I can't trigger the readthedocs build. If someone with permissions is able to activate this branch on readthedocs so I can verify the build, it would be appreciated (@denis-yuen maybe?)

:eyes:

denis-yuen:
Activated as a hidden version at https://docs.dockstore.org/en/feature-check-in-dictionary/getting-started/getting-started-with-notebooks.html 
It seems to work

svonworl (approved):
Good catch!

coverbeck (approved):
🤦 Thanks for catching and fixing my bug!

PR #289, opened 2024-08-15
Add baseurl from environment

david4096:
**Description**

RTD is deprecating the old way of getting the baseurl. This is the change they require to support running RTD on a custom domain. Now uses RTD addons

https://about.readthedocs.com/blog/2024/07/addons-by-default/

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-6573


coverbeck (approved):
We could just release the hotfix right away, making this PR moot, but this works too.

denis-yuen (approved):

PR #291, opened 2024-08-21
update sphinx comparison

denis-yuen:
**Description**
Update sphinx and transitive dependencies 
Mainly going from sphinx 4.4.0 (Jan 2022) to 7.4.7 (Jul 2024) 

**Issue**
See https://ucsc-cgl.atlassian.net/browse/SEAB-6642 


denis-yuen:
Ah, 
> Ignored the following versions that require a different python version: 1.0.0 Requires-Python >=3.10; 8.0.0 Requires-Python >=3.10; 8.0.0rc1 Requires-Python >=3.10; 8.0.1 Requires-Python >=3.10; 8.0.2 Requires-Python >=3.10

denis-yuen:
Oh yeah, I used `pip-compile` as documented at https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#pin-your-transitive-dependencies

aofarrel:
Updating these dependencies in particular can be tricky as it sometimes silently breaks formatting that can get missed at first glance. Will need to check the following before merging:
* pages generated from markdown (.md)
* pages with bullet points <-- this one in particular caused issues previously
* internal and external links
* the clickable images on the homepage
* tables

denis-yuen:
Thanks for the pointers, will look for those. 
Think we'll keep this open till early in 1.17 so we have some extra time to look

denis-yuen:
> Updating these dependencies in particular can be tricky as it sometimes silently breaks formatting that can get missed at first glance. Will need to check the following before merging:
> 
>     * pages generated from markdown (.md)
> 
>     * pages with bullet points <-- this one in particular caused issues previously
> 
>     * internal and external links
> 
>     * the clickable images on the homepage
> 
>     * tables
> 
> 
> If you run into any issues with glossarpy, ping me so I can make sure it remains up-to-spec!

I checked these items, particularly the tables, homepage, glossary, and the larger pages with bullet points. 
No issues noticed surprisingly. 

Things just seem the same comparing https://docs.dockstore.org/en/feature-experiment/ with https://docs.dockstore.org/en/stable/

docs/docker_instructions.md
```diff
@@ -6,7 +6,7 @@ The Docker CLI is a command-line tool with a whole library of commands for inter
 
 
 ### Part A - Running Containers
-In this section you will try running a basic container called whalesay. Refer to the [part A readings](#part-a-readings) if you need a refresher on the content taught in the training.
+In this section you will try running a basic container called whalesay. Refer to part A below if you need a refresher on the content taught in the training.
```
svonworl, line 9: What's the motivation for this change?  We're in "Part A" already, and the other "Part A" is a long way down the page (past another "Part A"), so the link is helpful.

denis-yuen: The immediate motivation is something about this method of linking in a md (vs rst) document is killing the build hard.  Presumably some dependency has changed it's behaviour. 

Exact error message is 
```
/root/project/docs/docker_instructions.md:9: WARNING: 'myst' cross-reference target not found: 'part-a-readings' [myst.xref_missing]
/root/project/docs/docker_instructions.md:39: WARNING: 'myst' cross-reference target not found: 'part-b-readings' [myst.xref_missing]
```

I'll try to describe it more.

denis-yuen: Ok, found the breaking change

svonworl: yay

svonworl (approved):

coverbeck (approved):

PR #294, opened 2024-11-07
SEAB-6738: Tweak Automatic DOI Section

svonworl:
**Description**
This PR enhances the new "Automatic Dockstore DOI Generation" section.  It is an attempt to codify the bits of information present in this Slack thread: https://ucsc-gi.slack.com/archives/C05EZH3RVNY/p1730923309122769

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6738

### DIFF
--- a/docs/end-user-topics/dois.rst
+++ b/docs/end-user-topics/dois.rst
@@ -69,10 +69,15 @@ Automatic Dockstore DOI Generation
 .. note::
     This feature is currently only available on request. Please contact the Dockstore team at `discuss.dockstore.org <https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506>`__ if you want to participate.
 
-* Automatically mints DOIs for all Git tags pushed to GitHub repositories that have a corresponding published entry in Dockstore
-* The DOIs are minted in the Zenodo Dockstore community
-* No Zenodo account is required to mint DOIs
-* A linked Zenodo account is required to edit DOIs
+Dockstore will automatically mint a DOI for a version that corresponds to a Git tag, if:
+
+* The entry is published on Dockstore
+* The version has at least one author in its metadata
+* The tag was created after the entry was published, or the tag is one of the 10 most recent
+
+* Automatic DOIs are minted in the Zenodo Dockstore community
+* No Zenodo account is required to mint automatic DOIs
+* In a subsequent release, automatic DOIs will be editable via a linked Zenodo account
 
 GitHub-Zenodo Generation
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/end-user-topics/dois.rst
+++ b/docs/end-user-topics/dois.rst
@@ -72,7 +72,7 @@ Automatic Dockstore DOI Generation
 Dockstore will automatically mint a DOI for a version that corresponds to a Git tag, if:
 
 * The entry is published on Dockstore
-* The version has at least one author in its metadata
+* The version is valid and has at least one author in its metadata
 * The tag is created after you publish the entry, or the tag is one of the 10 most recent when you publish the entry
 
 Automatically-generated DOIs:
--- a/docs/end-user-topics/dois.rst
+++ b/docs/end-user-topics/dois.rst
@@ -73,14 +73,14 @@ Dockstore will automatically mint a DOI for a version that corresponds to a Git 
 
 * The entry is published on Dockstore
 * The version has at least one author in its metadata
-* The tag was created after the entry was published, or the tag is one of the 10 most recent
+* The tag is created after you publish the entry, or the tag is one of the 10 most recent when you publish the entry
 
 Automatically-generated DOIs:
 
 * Do not require a Zenodo account
 * Are minted in the Zenodo Dockstore community
 * Are editable by request to the Dockstore team
-* Will be editable via a linked Zenodo account in a subsequent Dockstore release
+* Will be editable via a Zenodo account in a subsequent Dockstore release
 
 GitHub-Zenodo Generation
 ~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/end-user-topics/dois.rst
```diff
@@ -69,10 +69,17 @@ Automatic Dockstore DOI Generation
...
+
+* The entry is published on Dockstore
+* The version has at least one author in its metadata
+* The tag was created after the entry was published, or the tag is one of the 10 most recent
+
+Automatically-generated DOIs:
+
+* Do not require a Zenodo account
+* Are minted in the Zenodo Dockstore community
+* Will be editable via a linked Zenodo account in a subsequent Dockstore release
```
denis-yuen, line 82: Or by request to the Dockstore team

docs/end-user-topics/dois.rst
```diff
@@ -69,10 +69,18 @@ Automatic Dockstore DOI Generation
 .. note::
     This feature is currently only available on request. Please contact the Dockstore team at `discuss.dockstore.org <https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506>`__ if you want to participate.
 
-* Automatically mints DOIs for all Git tags pushed to GitHub repositories that have a corresponding published entry in Dockstore
-* The DOIs are minted in the Zenodo Dockstore community
-* No Zenodo account is required to mint DOIs
-* A linked Zenodo account is required to edit DOIs
+Dockstore will automatically mint a DOI for a version that corresponds to a Git tag, if:
+
+* The entry is published on Dockstore
```
coverbeck, line 74: Do you mean if the entry is already published, or the act of publishing? I'd read it as the former, but then looking at line 76, I think you mean the act of publishing. If so, I suggest something like "Publishing an entry on Dockstore".

docs/end-user-topics/dois.rst
```diff
@@ -69,10 +69,18 @@ Automatic Dockstore DOI Generation
...
+* The version has at least one author in its metadata
+* The tag was created after the entry was published, or the tag is one of the 10 most recent
+
+Automatically-generated DOIs:
+
+* Do not require a Zenodo account
+* Are minted in the Zenodo Dockstore community
+* Are editable by request to the Dockstore team
+* Will be editable via a linked Zenodo account in a subsequent Dockstore release
```
coverbeck, line 83: @kathy-t Should confirm, but I don't think the Zenodo account will need to be linked to Dockstore. I could be wrong, but I thought this feature will generate a signed Zenodo URL that anybody in possession of can edit with a Zenodo account.

coverbeck (approved):

docs/end-user-topics/dois.rst
svonworl: I removed "linked", we can add it back later if it turns out that it's accurate and important to note...

docs/end-user-topics/dois.rst
svonworl: > Do you mean if the entry is already published, or the act of publishing? I'd read it as the former, but then looking at line 76, I think you mean the act of publishing. If so, I suggest something like "Publishing an entry on Dockstore".

Both meanings. I changed the last bullet to clarify the two cases a little better.
Might not describe all of the scenarios 100% precisely, but short and readable and more future proof than trying to quantify the retroactive DOI generation that we'll probably do later but don't right now...

docs/end-user-topics/dois.rst
```diff
@@ -69,10 +69,18 @@ Automatic Dockstore DOI Generation
...
     This feature is currently only available on request. Please contact the Dockstore team at `discuss.dockstore.org <https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506>`__ if you want to participate.
 
-* Automatically mints DOIs for all Git tags pushed to GitHub repositories that have a corresponding published entry in Dockstore
-* The DOIs are minted in the Zenodo Dockstore community
-* No Zenodo account is required to mint DOIs
-* A linked Zenodo account is required to edit DOIs
+Dockstore will automatically mint a DOI for a version that corresponds to a Git tag, if:
+
+* The entry is published on Dockstore
+* The version has at least one author in its metadata
```
kathy-t, line 75: The version also has to be valid https://github.com/dockstore/dockstore/blob/d8e12f97434dc60437e8e6bfffbec5ac00da798b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java#L965

docs/end-user-topics/dois.rst
kathy-t: Yeah that's correct, the user won't need to link their account to Dockstore, but they will need to sign into Zenodo after they click the secret Zenodo URL in order to edit the DOI record

docs/end-user-topics/dois.rst
kathy-t, line 1: should we add a reference to this page from https://github.com/dockstore/dockstore-documentation/blob/develop/docs/advanced-topics/snapshot-and-doi.rst so users can see that there are other options?

svonworl: > should we add a reference to this page from https://github.com/dockstore/dockstore-documentation/blob/develop/docs/advanced-topics/snapshot-and-doi.rst so users can see that there are other options?

Good idea, but would probably suggest that we do it when the Automatic DOI feature is out of "preview".

david4096 (approved):

docs/end-user-topics/dois.rst
```diff
@@ -69,10 +69,18 @@ Automatic Dockstore DOI Generation
...
+
+* The entry is published on Dockstore
+* The version is valid and has at least one author in its metadata
+* The tag is created after you publish the entry, or the tag is one of the 10 most recent when you publish the entry
+
+Automatically-generated DOIs:
+
+* Do not require a Zenodo account
+* Are minted in the Zenodo Dockstore community
+* Are editable by request to the Dockstore team
```
denis-yuen, line 82: To be clear, we probably won't do the editing but can attach their account to the entry in zenodo and let them edit it, sort of a manual version of the next point

denis-yuen (approved):
Minor clarification, not sure if worth pointing out

kathy-t (approved):

PR #296, opened 2024-12-13
document automatic dois

denis-yuen:
**Description**
A description of the PR, should include a decent explanation as to why this change was needed and a decent explanation as to what this change does

**Issue**
A link to a github issue or SEAB- ticket (using that as a prefix)

- [ ] Ensure that the PR targets the correct branch. Check the milestone or fix version of the ticket.

### DIFF
--- a/docs/assets/templates/notebooks/template-big.dockstore.yml
+++ b/docs/assets/templates/notebooks/template-big.dockstore.yml
@@ -63,6 +63,11 @@ notebooks:
     # A value of true will automatically display the latest tag updated as default.
     # A value of false will retain the default version that has been specified via the Dockstore UI.
     latestTagAsDefault: <Boolean>
+
+    # A boolean that will control whether DOIs are automatically issued for tags. Default: True.
+    # A value of true will allow Dockstore to automatically issue DOIs.
+    # A value of false will disable automatic DOIs, DOIs will only be issued on demand.
+    enableAutoDois: <Boolean>
     
     # The optional filters section allow specifying sets of Git branches and tags to include for the notebook.
     # If no filters are given, all branches and tags are included.
--- a/docs/end-user-topics/dois.rst
+++ b/docs/end-user-topics/dois.rst
@@ -107,6 +107,12 @@ Automatically-generated DOIs:
 * Are editable by request to the Dockstore team
 * Will be editable via a Zenodo account in a subsequent Dockstore release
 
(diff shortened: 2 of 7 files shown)


denis-yuen:
Note to self: should document datacite automatic orcid addition issue

svonworl (approved):

docs/assets/templates/notebooks/template-big.dockstore.yml
```diff
@@ -63,6 +63,11 @@ notebooks:
     # A value of true will automatically display the latest tag updated as default.
     # A value of false will retain the default version that has been specified via the Dockstore UI.
     latestTagAsDefault: <Boolean>
+
+    # A boolean that will control whether DOIs are automatically issued for tags. Default: True.
```
kathy-t, line 67: Setting it to `true` right now will cause the push to fail. The only acceptable value is `false` since we control the enabling via the admin endpoint

kathy-t, line 67: Suggest adding the following, similar to the `publish` and `topic` properties above

>  Workflow-wide setting that will affect ALL branches/tags; only set this as needed in a main branch.

docs/end-user-topics/dois.rst
```diff
@@ -107,6 +107,18 @@ Automatically-generated DOIs:
...
+
+.. code-block:: yaml
+
+   enableAutoDois: false
+
+
+One rare case to potentially watch out for: if your workflows are tagged with your ORCID as an author
+you may get a large number of notifications or workflows automatically added to your ORCID profile.
+This will occur outside of Dockstore if you have the `ORCID Auto-Update <https://support.datacite.org/docs/datacite-and-orcid#2-orcid-auto-update>`__ feature turned on in DataCite.
+DataCite's auto-update feature is not aware of Zenodo's concept DOI concept and will import each version independently.
```
kathy-t, line 120: "Zenodo's concept DOI concept" reads a bit awkwardly to me

kathy-t (commented):
Should this target the hotfix branch?

denis-yuen:
> Should this target the hotfix branch?

yeah, good point

docs/assets/templates/notebooks/template-big.dockstore.yml
denis-yuen: Hmmmm, actually fail as in aborting or just fail to do anything?

denis-yuen:
closing in favour of #297

PR #297, opened 2024-12-16
document auto dois opt-out setting

denis-yuen:
**Description**
Document DOI opt-out option in templates and in the DOI article. 

View draft at https://docs.dockstore.org/en/feature-document_enable_auto_doi/

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6833

### DIFF
--- a/docs/assets/templates/workflows/template-big.dockstore.yml
+++ b/docs/assets/templates/workflows/template-big.dockstore.yml
@@ -65,6 +65,11 @@ workflows:
     # A value of true will automatically display the latest tag updated as default.
     # A value of false will retain the default version that has been specified via the Dockstore UI.
     latestTagAsDefault: <Boolean>
+
+    # A boolean that will control whether DOIs are automatically issued for tags. Default: True.
+    # A value of true will allow Dockstore to automatically issue DOIs.
+    # A value of false will disable automatic DOIs, DOIs will only be issued on demand.
+    enableAutoDois: <Boolean>
     
     # The optional filters section allow specifying sets of Git branches and tags to include for the workflow.
     # If no filters are given, all branches and tags are included.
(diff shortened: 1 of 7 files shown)


docs/end-user-topics/dois.rst
```diff
@@ -82,6 +82,18 @@ Automatically-generated DOIs:
 * Are editable by request to the Dockstore team
 * Will be editable via a Zenodo account in a subsequent Dockstore release
 
+If you wish to override and disable this feature, you can opt-out on a per-workflow basis you can add the following setting to your `.dockstore.yml` files
```
kathy-t, line 85: What comes after the comma sounds like it's missing something. Maybe 
> you can opt-out on a per-workflow basis by adding the following setting to your `.dockstore.yml` files

kathy-t (approved):

docs/assets/templates/workflows/template-big.dockstore.yml
```diff
@@ -65,6 +65,11 @@ workflows:
     # A value of true will automatically display the latest tag updated as default.
     # A value of false will retain the default version that has been specified via the Dockstore UI.
     latestTagAsDefault: <Boolean>
+
+    # 'enableAutoDois' is a workflow-wide setting that will affect ALL branches/tags; only set this as needed in a main branch
```
coverbeck, line 69: Beyond the scope of this PR, but this statement got me thinking, maybe the web service should only honor this and other workflow-level scoped settings if on the GitHub default branch.

Even with this advice, if a user reads it, it's going to be hard to do; if you create a new branch from the default branch, the .dockstore.yml is copied to the new branch.

No need to change anything here; I'll create a ticket for this idea; not sure if it's feasible.

coverbeck (approved):

docs/assets/templates/workflows/template-big.dockstore.yml
coverbeck: https://ucsc-cgl.atlassian.net/browse/SEAB-6869

svonworl (approved):

PR #302, opened 2025-03-25
docker schema update

denis-yuen:
**Description**
Follow-up for Docker schema deprecation issue https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1  We've switched integration tests before but not the tutorials. 
Test that Docker images used in tutorials are monitored by github in case of future deprecation
Updated container that we control, switched when we cannot. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6852

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,22 @@
+name: Docker Image CI
+
+on: [push]
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    # Check on tutorial docker containers 
+    - run: docker pull quay.io/briandoconnor/dockstore-tool-md5sum
+    - run: docker pull docker/whalesay
+    - run: docker pull quay.io/ldcabansay/samtools
+    - run: docker pull tabix:latest
+    - run: docker pull quay.io/collaboratory/dockstore-tool-bamstats:1.25-6
+    - run: docker pull pkrusche/hap.py:v1.0
+    - run: docker pull pkrusche/hap.py:latest
+    - run: docker pull biocontainers/samtools:v1.9-4-deb_cv1
+    - run: docker pull pkrusche/hap.py:latest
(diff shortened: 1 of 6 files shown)


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,23 @@
...
+on: [push]
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
```
kathy-t: I think this action can be removed because the job isn't using anything from the repo

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,23 @@
...
+    # Check on tutorial docker containers 
+    - run: docker pull quay.io/dockstore-testing/dockstore-tool-md5sum
+    - run: docker pull rancher/cowsay
+    - run: docker pull quay.io/ldcabansay/samtools
+    # this container is built by the user, not pulled from docker hub
+    # - run: docker pull tabix:latest
```
kathy-t, line 18: Do we need to keep these commented out lines? If yes, some comments about why would be helpful

kathy-t (approved):

docs/docker_instructions.md
```diff
@@ -115,7 +108,7 @@ There are readings to help with the exercises. Refer here for any questions you
 
 ### Exercise 1
 #### Part A Readings
-Whalesay is a program that given some text, will print out an ASCII whale that is saying the text. It is based on a program called cowsay.
+Cowsay is a program that given some text, will print out an ASCII cow that is saying the text. It is based on a program called cowsay.
```
svonworl, line 111: There's several lingering "whale" references in this file, `grep -rIi whale` will find them

.github/workflows/docker-image.yml
denis-yuen: Kinda. I found these in a text search and they look like containers. Will elaborate

svonworl (approved):

